PR #9, opened 2016-01-22
feat(plugin): add globbing support for from option (fixes #7)

### LINKED ISSUE #7: Globs for `from`
Would be much easier to follow than inverted `exclude` patterns.

mgcrea:
Tried to avoid any breaking change for now.

In my opinion, it could be worth it to release a new breaking major after that only deals with globbing with both `cwd` and `base` options (like gulp/vinyl).

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 var _ = require('lodash');
 var Promise = require('bluebird');
 var path = require('path');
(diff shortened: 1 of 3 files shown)


kevlened:
I like it! I'll bring it in. I'm going to hold off on the cwd and base globbing for now

Issue #15, opened 2016-01-29
[Feature] copy files after webpack has finished creating the bundles


jmlopez-rod:
I was hoping that I could use this plugin to create copies of the generated bundles to other directories. It seems however that the copying starts while webpack is creating these bundles. As a result I end up with copies of the files with no content. 

Is there a way to add an option so that the copying of the files happen after webpack has created the output files?

jmlopez-rod:
Ok, so after digging around I'm starting to believe that there is no easy way to do this with the current version of the plugin.  In `writeFileToAssets` the file source gets attached to `compilation.assets`. 

```js
        compilation.assets[relFileDest] = {
          size: function() {
            return stat.size;
          },
          source: function() {
            return fs.readFileSync(absFileSrc);
          }
        };
```

The problem is, that the source may not yet be there since webpack hasn't written it to the file yet so the source will be empty. So the next thing to try would be, can we check if `absFileSrc` is in the assets and make it read that source? I'll post when I find out.

jmlopez-rod:
So here is a solution, I added a few lines to `writeFileToAssets` so that it can check if the file is in production then to just point to that asset.

Here is the whole function

``` js
function writeFileToAssets(opts) {
  var compilation = opts.compilation;
  var relFileDest = opts.relFileDest;
  var absFileSrc = opts.absFileSrc;
  var forceWrite = opts.forceWrite;
  var lastGlobalUpdate = opts.lastGlobalUpdate;

  if (compilation.assets[relFileDest] && !forceWrite) {
    return Promise.resolve();
  }

  var currentAssets = _.keys(compilation.assets);
  var inProduction = _.find(currentAssets, function(name) {
    return _.endsWith(absFileSrc, name);
  });

  return fs.statAsync(absFileSrc)
    .then(function(stat) {
      if (stat.mtime.getTime() > lastGlobalUpdate) {
        if (compilation.assets[inProduction]) {
          compilation.assets[relFileDest] = compilation.assets[inProduction];
        } else {
          compilation.assets[relFileDest] = {
            size: function() {
              return stat.size;
            },
            source: function() {
              return fs.readFileSync(absFileSrc);
            }
          };
        }
      }
    });
}
```

Would you mind incorporating this change?

iamakulov:
Seems like this works now.

tobinibot:
[...]

PR #21, opened 2016-02-08
copy files in webpack stream: Issue #15

jmlopez-rod:
Checking if the file we are trying to copy is one of the current assets
in webpack. If this is the case then we set the source of file to copy
to the one already in the assets. Otherwise we proceed as before.


kevlened:
We currently ignore files that are already in `compilation.assets` unless the `force` option is passed, so I'm not quite sure what you're proposing

PR #23, opened 2016-02-15
Support lodash v4.0.0

adamsxm:
This adds support for lodash 4.0 which has breaking changes and replaces `_.contains` with `_.includes` per [Changelog](https://github.com/lodash/lodash/wiki/Changelog#jan-12-2016--diff--docs)


mgeduld:
Thank you!!!!

PR #27, opened 2016-02-24
Add the possibility to define a pattern specific context

chrigu-ebert:
Hi there,

For one of my projects I needed to be able to specify a pattern specific context and this is why I added that new optional parameter.

My usecase is the following:
I want to copy html files from the app directory but at the same time copy fonts from one of my vendor node_modules folder.

Example:

`new CopyWebpackPlugin([
      {
        context: path.join(__dirname, 'app'),
        from: "**/*.html"
      },
      {
        context: path.join(__dirname, 'node_modules/ionic-framework'),
        from: "fonts/*.ttf"
      }
    ])`

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,10 @@ A pattern looks like:
     - is optional
     - defaults to `false`
     - forces the plugin to overwrite files staged by previous plugins
+* `context`
+    - is optional
+    - is a pattern specific context
+    - overrides the base context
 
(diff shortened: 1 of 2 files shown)


kevlened:
I like this idea! It'd certainly help with globbing use cases too.

kevlened:
This has been added in v2.1.0. Thanks!

PR #33, opened 2016-04-07
remove platform specific separators from compilation.asset keys, fixes #29

### LINKED ISSUE #29: Does not copy files to actual output folder when webpack-dev-server is used
It appears that nothing happens when trying to copy some extra files (not bundled) to the output folder, under the `webpack-dev-server` logic:

````
devServer: {
            contentBase: './public',
            outputPath: path.resolve(ROOT_PATH, 'public'),
            historyApiFallback: true,
            hot: true,
            inline: true,
            port: 33000,
            progress: true
        },
        plugins: [
            new CopyWebpackPlugin([
                {from: 'src/css', to: 'css', force: true},
                {from: 'src/img', to: 'img', force: true}
            ]),
            new webpack.HotModuleReplacementPlugin(),
            new OpenBrowserPlugin({url: 'http://localhost:33000'})
        ]
````

Can this be done somehow? The webpack-dev-server needs this to show some external CSS and IMAGE files. I used to be doing this with `gulp` but now I want to get rid of this dependency.

Using your plugin with the simple webpack (for `build`) works perfectly but not for `webpack-dev-server`.

NOTE: I can achieve what I want if I use both `webpack` and `webpack-dev-server` on the `start` script, but this is not too elegant and done only for solving this problem:
````
  "scripts": {
    "build": "webpack",
    "start": "npm run build && webpack-dev-server"
  },
````

yntelectual:
Proposed PR to fix the compilation.asset keys handling for paths on windows as described in #29.

### DIFF
--- a/index.js
+++ b/index.js
@@ -138,6 +138,8 @@ function writeFileToAssets(opts) {
   var absFileSrc = opts.absFileSrc;
   var forceWrite = opts.forceWrite;
   var lastGlobalUpdate = opts.lastGlobalUpdate;
+  //make sure that asset keys use fwd slashes
+  var relFileDest = relFileDest.replace(/\\/g, "/");
 
   if (compilation.assets[relFileDest] && !forceWrite) {
     return Promise.resolve();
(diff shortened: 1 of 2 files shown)


kevlened:
Thanks @yntelectual ! There was a massive change to ES6, so I applied your change manually. This is now fixed in v2.0.1.

PR #37, opened 2016-04-21
Fix TypeError when working with webpack-dev-server

tommytroylin:
webpack version 1.12.12 + webpack-dev-server 1.14.1

Seems like `compiler.options.devServer` is **undefined** when user doesn't defined `devServer` in webpack config

it will throw `TypeError: Cannot read property 'outputPath' of undefined`  at function `getOutputDir()`

So the customized error won't be thrown out

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -24,13 +24,13 @@ const getOutputDir = (compiler) => {
         return compiler.options.output.path;
     }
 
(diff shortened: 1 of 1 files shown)


kevlened:
lgtm. Thanks!

PR #40, opened 2016-05-03
Fix ERROR in Path must be a string. Received undefined

tommytroylin:
node 6.0.0 + webpack 1.12.12

to reproduce this error, simply provide a file path in `from ` and omit `to`

like this in webpack config
`plugins: [
    new CopyWebpackPlugin([{ from: 'src/static/index.html' }]),
  ],`


kevlened:
Thanks! I'll have to investigate more to see why our tests didn't catch this

PR #43, opened 2016-05-07
Fixed writing to memory fs if webpack-dev-server

pfrankov:


kevlened:
Writing to the filesystem is a feature. Currently, we write it to the memory fs as well. What's the reason you don't want to write to the filesystem?

If it's necessary, we can include an option to disable it

pfrankov:
@kevlened I don't see any reason to want to write to the fs in the dev mode when all other files are in the memory fs.

kevlened:
I'm no longer writing to the filesystem directly in v3.0.0. If anyone needs to write to the filesystem while the devServer is running, they should include the [write-file-webpack-plugin](https://github.com/gajus/write-file-webpack-plugin).

PR #62, opened 2016-07-05
add contenthash options

lcxfs1991:
We may need content hash for copied files too.

I run the tests and it shows the new feature does not affect old feature.

However, I don't have time to write a test for this feature, would u please write it for me?

Thank u so much!


ksmithbaylor:
I'd also love to see this feature implemented, and would be willing to help out.

kevlened:
Sorry for the delay on this. After some investigation, it seems that `loader-utils` has this functionality built in with `interpolateName`. I wasn't aware of loader-utils before this PR, so thanks! The master branch (not yet released) has template support.

evilebottnawi:
@kevlened be good release with new versions in package.json

kevlened:
This is now supported in version 4.0.0, and published to npm. Thanks again!

lcxfs1991:
@kevlened can't locate the details in README. Haven't been updated?

kevlened:
It's under https://github.com/kevlened/copy-webpack-plugin#usage. If you're looking for all the template options (including hash), the details of every template we support is here: https://github.com/webpack/file-loader#filename-template-placeholders

There's no `[contenthash]`, but `[hash]` should be what you need

Issue #84, opened 2016-10-26
copy folders in Windows


padinko:
I am using Windows 7,
I have folder with subfolder which i want copy

problem is with subfolder and dev-server: this plugin return assets with backslash and dev-server don't find this files

example:
`assets\lang\en.json` from plugins
`assets/lang/en.json` from browser request

larsthorup:
I have the same problem (win10, node@7, copy-webpack-plugin@4.0.0).

Running the tests with `npm run build && npm run build:tests && mocha compiled_tests/` gives 30 failing tests, all involving mismatched slashes and backslashes, like e.g.

    1) apply function with glob in from can use a glob to move multiple files to the root directory:

      AssertionError: expected { Object (binextension.bin, file.txt, ...) } to have keys 'binextension.bin', 'file.txt', 'directory/directoryfile.txt', 'directory/nested/nestedfile.txt', and 'noextension'
      + expected - actual

       [
         "binextension.bin"
      -  "directory\\directoryfile.txt"
      -  "directory\\nested\\nestedfile.txt"
      +  "directory/directoryfile.txt"
      +  "directory/nested/nestedfile.txt"
         "file.txt"
         "noextension"
       ]

      at C:\r\github\copy-webpack-plugin\compiled_tests\index.js:117:67

Downgrading to copy-webpack-plugin@3.0.1 fixes the issue for me for now.

larsthorup:
I will see if I can figure out a fix for this...

kevlened:
@larsthorup Saw your PR; I'll get that in soon. This is definitely a gap in our test coverage. I think I can add tests around this without spinning up a cross-platform CI.

PR #85, opened 2016-10-31
improved windows compatibility

larsthorup:
improved windows compatibility, in reference to #84

most tests now pass on windows. the code line added was taken from the original writeFileToAssets.js, line 14


schnittstabil:
Would be nice to have some CI for that, like https://ci.appveyor.com provides (free for OSS).
Similar to travis, appveyor needs a configuration file (`appveyor.yml`, an advanced example: https://github.com/avajs/ava/blob/master/appveyor.yml).

PR #91, opened 2016-11-10
feat: add writing of cache file and loading it

### LINKED ISSUE #90: [Feature] Persisting between full webpack builds
You do have `copyUnmodified` set to `false` as default, but that cache is never stored anywhere and thus fails to actually cache anything

deiga:
### Notable Changes
---

Only problem here, is that most tests fail, because we skip files and they are not addded to compilation assets.

A solution would be to change the logic here, so that files are always added to compilation assets.

Another solution could be to skip writing cache files and just compare `from` and `to` files with each other to see if there are cahnges

### Issues
---

- Fixes #90 

> ℹ️  edited by @michael-ciniawsky (Formatting)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,16 +6,17 @@
   "license": "MIT",
   "main": "dist/index.js",
   "dependencies": {
-    "bluebird": "^2.10.2",
-    "find-cache-dir": "^0.1.1",
-    "fs-extra": "^0.26.4",
-    "glob": "^6.0.4",
+    "bluebird": "^3.5.1",
+    "find-cache-dir": "^1.0.0",
+    "fs-extra": "^4.0.2",
+    "glob": "^7.1.2",
+    "is-glob": "^4.0.0",
     "loader-utils": "^0.2.15",
     "lodash": "^4.3.0",
-    "minimatch": "^3.0.0",
+    "minimatch": "^3.0.4",
     "node-dir": "^0.1.10"
   },
-  "homepage": "https://github.com/kevlened/copy-webpack-plugin",
+  "homepage": "https://github.com/webpack-contrib/copy-webpack-plugin",
   "scripts": {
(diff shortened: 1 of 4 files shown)


michael-ciniawsky:
@deiga Could you please rebase against current `master` ? :)

package.json
```diff
@@ -24,6 +16,16 @@
     "minimatch": "^3.0.4",
     "node-dir": "^0.1.10"
   },
+  "homepage": "https://github.com/webpack-contrib/copy-webpack-plugin",
```
evilebottnawi: Don't change anything in `package.json`, only add needed package(s), thank you!

src/index.js
```diff
@@ -131,4 +149,4 @@ function CopyWebpackPlugin(patterns = [], options = {}) {
 }
 
 CopyWebpackPlugin['default'] = CopyWebpackPlugin;
-module.exports = CopyWebpackPlugin;
+module.exports = CopyWebpackPlugin;
```
evilebottnawi: Don't do this

src/writeFile.js
```diff
@@ -73,6 +73,9 @@ export default function writeFile(globalRef, pattern, file) {
                 written[file.absoluteFrom] = {
                     [hash]: true
                 };
+                if (!copyUnmodified) {
+                    fs.writeFileSync(cachePath, JSON.stringify(written));
```
evilebottnawi: Sync?

deiga:
@evilebottnawi You're most likely right :P

Not even sure if this makes much sense to implement at all.
Some planning with other people would likely be a good idea.

evilebottnawi:
@deiga i can do `cache` after some basic changes in this repo, thank you, can we close this here?

deiga:
Sure, go ahead

PR #115, opened 2017-02-12
feat: add support for target path transforms (`options.transformPath`)

### LINKED ISSUE #107: [Feature] Manually modify output path
Hi!

It would be great if something like this would be supported:

```javascript
{
  context: './src/',
  from: '**/i18n/*',
  to: function(url) {
    return path.resolve('i18n', url.replace('i18n/', ''));
  }
}
```
Basically, we want to get all localization files and output them under a single i18n folder, preserving folder structure but without i18n.

I can also try adding this via PR during next days, if anyone else is interested.

patkec:
Added a new function, `transformPath`, with which the target path can be modified programmatically. Decided to keep the functionality of `to` param unmodified so that path transfformation can be combined with templates for ease of use.
Should resolve #107.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ A pattern looks like:
 | `flatten` | N | false | Removes all directory references and only copies file names<br><br>If files have the same name, the result is non-deterministic |
 | `ignore` | N | [] | Additional globs to ignore for this pattern |
 | `transform` | N | function(content, path) {<br>&nbsp;&nbsp;return content;<br>} | Function that modifies file contents before writing to webpack |
+| `transformPath` | N | function(targetPath, sourcePath) {<br>&nbsp;&nbsp;return targetPath;<br>} | Function that modifies target file path before writing to webpack |
 | `force` | N | false | Overwrites files already in compilation.assets (usually added by other plugins) |
 
 #### Available options:
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -64,6 +64,10 @@ export default function writeFile(globalRef, pattern, file) {
                 }
             }
 
+            if (typeof pattern.transformPath === 'function') {
+                file.webpackTo = pattern.transformPath(file.webpackTo, file.absoluteFrom);
+            }
+
             if (!copyUnmodified &&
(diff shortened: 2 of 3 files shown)


michael-ciniawsky:
@patkec Could you please rebase to latest `master` and without looking at the current implementation I wonder if it would be possible to accept a `{Function}` on already existing options/interfaces instead of adding a new option ?

patkec:
Rebased. As for the implementation, the only option that would be suitable is `to`, but then we would lose the possibility to use templates. I thought it nice that users could specify a template in `to` and still do some path processing after.

michael-ciniawsky:
Can you give a small example for `to`  and the loss of templates ? The implementation is fine, but we need to carefully triage the API design, since every added option is likely to stay for a while :)

patkec:
It seemed like a good idea at the time, but cannot really think of a good example right now. Will think about it over the weekend and get back to you.

patkec:
Thinking a bit about this, the only benefit of having both templates and a transform is that templates support content hashing. Everything else should be fairly easy to implement in the transform function, if needed.
No idea how often this hash option is used and if it is worth introducing a special transformPath option because of it.

michael-ciniawsky:
@evilebottnawi We should find consensus on this if possible. I'm -0 atm

FireController1847:
If you want, I can provide an example usage for this.

Let's say I want to copy all of my php files to my dist folder. When I do this currently:
```js
    new CopyWebpackPlugin([
      { from: 'src/*.html',
        to: './'
    ], { debug: 'info' })
```
This will make a new folder in `dist` called `src`, and THEN put my html files in that. I could use this by doing something like this:
```js
    new CopyWebpackPlugin([
      { from: 'src/*.html',
        to: './',
        transform: function(content, filePath) {
          return path.resolve(filePath, '..');
        } }
    ], { debug: 'info' })
```
But currently this does not work since we can't transform the path of a file, YET.

kdelmonte:
Any idea when this will be released?

evilebottnawi:
Need rebase

evilebottnawi:
Close in favor https://github.com/webpack-contrib/copy-webpack-plugin/pull/284

PR #119, opened 2017-03-14
feat: add support for copying file permissions (`options.copyPermissions`)

### LINKED ISSUE #35: Plugin is not copying file modes with execute permissions
We have cordova hooks in our source folder (see the cordova hooks guide for more info: https://cordova.apache.org/docs/en/dev/guide/appdev/hooks/) that need to be copied into our destination folder and need to preserve the executable bit on their file mode (i.e. `ls -l` should print `-rwxr-xr-x` for these files).

Currently this doesn't happen, but it seems like it should be easy to do by passing a options object as a 3rd parameter to the to the `fs.writeFileAsync()` call in the `writeFilePromises` with the `stat.mode` from the `fs.statAsync()` call.

deryni:
Add support for copying permissions for copied files.

Use a `copyPermissions: true` key in a copy pattern to turn it on.

Should resolve #35.

### DIFF
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -80,6 +80,19 @@ export default function writeFile(globalRef, pattern, file) {
                 return;
             }
 
+            let perms;
+            if (pattern.copyPermissions) {
+                debug(`saving permissions for '${file.absoluteFrom}'`);
+
+                written[file.absoluteFrom].copyPermissions = pattern.copyPermissions;
+                written[file.absoluteFrom].webpackTo = file.webpackTo;
+
+                perms |= stat.mode & fs.constants.S_IRWXU;
+                perms |= stat.mode & fs.constants.S_IRWXG;
+                perms |= stat.mode & fs.constants.S_IRWXO;
+                written[file.absoluteFrom].perms = perms;
+            }
+
             info(`writing '${file.webpackTo}' to compilation assets from '${file.absoluteFrom}'`);
             compilation.assets[file.webpackTo] = {
                 size: function() {
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -3,6 +3,7 @@ import loaderUtils from 'loader-utils';
 import path from 'path';
 
 const fs = Promise.promisifyAll(require('fs')); // eslint-disable-line import/no-commonjs
+const constants = Promise.promisifyAll(require('constants')); // eslint-disable-line import/no-commonjs
 
 export default function writeFile(globalRef, pattern, file) {
     const {info, debug, compilation, fileDependencies, written, copyUnmodified} = globalRef;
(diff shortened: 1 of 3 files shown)


deryni:
I tried writing tests for this but ran into trouble getting the stats for the emitted files.

(I can't seem to find them post-webpack to get the stats from.)

Antrikshy:
I'm setting up a new stack for an internal project at work that relies on this plugin. I'd love to see this pulled because we're currently relying on a slightly unsightly workaround for files that need to be copied with permissions.

j0k3r:
Definitely needs this in case you want to use executable file inside your lambda :+1:

src/writeFile.js
```diff
@@ -80,6 +80,19 @@ export default function writeFile(globalRef, pattern, file) {
...
+            let perms;
+            if (pattern.copyPermissions) {
+                debug(`saving permissions for '${file.absoluteFrom}'`);
+
+                written[file.absoluteFrom].copyPermissions = pattern.copyPermissions;
+                written[file.absoluteFrom].webpackTo = file.webpackTo;
+
+                perms |= stat.mode & fs.constants.S_IRWXU;
+                perms |= stat.mode & fs.constants.S_IRWXG;
+                perms |= stat.mode & fs.constants.S_IRWXO;
```
j0k3r: These constants (`S_IRWXU`, `S_IRWXG`, `S_IRWXO`) doesn't exist in node < 6.3 :confused:

deryni: Do they exist in the `constants` module (i.e. `require("constants")` in the earlier versions?

j0k3r: Yep they do exist in `constants.default.S_IRWXU` for example

j0k3r: I've tested with:

```js
// src/writeFile.js
perms |= stat.mode & (fs.constants.S_IRWXU || constants.default.S_IRWXU);
```

and 

```js
// src/index.js
let mask
if (fs.constants !== undefined) {
  mask = fs.constants.S_IRWXU | fs.constants.S_IRWXG | fs.constants.S_IRWXO;
} else {
  mask = constants.default.S_IRWXU | constants.default.S_IRWXG | constants.default.S_IRWXO;
}
```

And it works fine on node 6.3

michael-ciniawsky:
Closing due to inactivity, feel free to reopen :)

j0k3r:
@michael-ciniawsky can't you retrieve its commits and do the rebase on your own?
We are using a fork with its modification and it works pretty good so far.

michael-ciniawsky:
@j0k3r Feel free to open a PR with these changes, this still needs review, e.g does this _really_ need to be an option or could it be a default instead etc ? :)

PR #134, opened 2017-06-27
refactor: updated ignore to allow path expressions when flatten is true

nrowan:
Previously the ignore would only be run against the file name in the "From" property. I added a test to show the case that was broken.

The below line essentially strips the path before the ignore statement and prevented the ability to ignore anything but the filename itself.
file.relativeFrom = path.basename(file.relativeFrom);


nrowan:
Is there anybody still working on this project anymore?

michael-ciniawsky:
@nrowan Yup, it's now under && maintained by webpack-contrib, could you please rebase so it is easier to review your proposed changes ? :)

michael-ciniawsky:
Closing due to inactivity, feel free to reopen :)

tinyfly:
Problem still exists. If I set flatten:true then I can't ignore based on "from" directory structure.

PR #136, opened 2017-07-13
feat: allow multiple outputs for same input

### LINKED ISSUE #99: [Bug] Copy multiple time the same file does one input only
I tried to do something like this:
```js
plugins: [
  new CopyWebpackPlugin([
    {
      from: 'a.js',
      to: 'b.js',
    },
    {
      from: 'a.js',
      to: 'c.js',
    },
  ]),
],
```

When doing that, only one output file is created. 
I had to do this to make it work:
```js
plugins: [
  new CopyWebpackPlugin([
    {
      from: 'a.js',
      to: 'b.js',
    },
  ]),
  new CopyWebpackPlugin([
    {
      from: 'a.js',
      to: 'c.js',
    },
  ]),
],
```

jdeniau:
Fixes #99

### DIFF
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -65,13 +65,18 @@ export default function writeFile(globalRef, pattern, file) {
             }
 
             if (!copyUnmodified &&
-                written[file.absoluteFrom] && written[file.absoluteFrom][hash]) {
+                written[file.absoluteFrom] &&
+                written[file.absoluteFrom][file.webpackTo] &&
+                written[file.absoluteFrom][file.webpackTo][hash]
+            ) {
                 info(`skipping '${file.webpackTo}', because it hasn't changed`);
                 return;
             } else {
                 debug(`added ${hash} to written tracking for '${file.absoluteFrom}'`);
                 written[file.absoluteFrom] = {
-                    [hash]: true
+                    [file.webpackTo]: {
+                        [hash]: true
+                    }
                 };
             }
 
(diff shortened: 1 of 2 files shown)


michael-ciniawsky:
@jdeniau Please rebase to latest `master`, so we can triage this further :)

jdeniau:
@michael-ciniawsky done

src/writeFile.js
```diff
@@ -65,13 +65,18 @@ export default function writeFile(globalRef, pattern, file) {
...
                 info(`skipping '${file.webpackTo}', because it hasn't changed`);
                 return;
             } else {
                 debug(`added ${hash} to written tracking for '${file.absoluteFrom}'`);
                 written[file.absoluteFrom] = {
-                    [hash]: true
+                    [file.webpackTo]: {
+                        [hash]: true
+                    }
                 };
```
d3viant0ne: Do you have a specific use case that can't be accomplished using `copyUnmodified`.

Based on the test case below, you could do the same thing without adding new functionality.

jdeniau: I improved the test to add a transform method to the plugin. This way it does not work even if you add `copyUnmodified` option.

For the record, when I had the bug, it was by copying the same file with the same transformation (basically babel + uglify) in two different locations for history purpose (two different files became only one, but both are referenced)

jdeniau:
For the record the test is failing in Node 4 only with a gcc bug, I think you can restart it and it should work.

Thank

d3viant0ne:
Triggered a rebuild on Travis.

evilebottnawi:
Duplicate https://github.com/webpack-contrib/copy-webpack-plugin/pull/165

PR #149, opened 2017-10-05
feat: add `context` option (`options.context`)

### LINKED ISSUE #148: [Feature] Provide default `context` option
Currently, `context` defaults to `compiler.options.context` from webpack. However, I'd like all my copy patterns to use `./src` as the default context, even though I can't change my webpack context to that for other reasons.

Could we add a fallback to the plugin options? The overall context fallback logic would be:

`context` property on pattern object -> `context` property in plugin options -> `context` property on webpack config

If this sounds okay, I can try to come up with a PR.

cletusw:
Overrides webpack context, overridden by pattern context

Fixes #148

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
 import Promise from 'bluebird';
+import path from 'path';
 import _ from 'lodash';
 import preProcessPattern from './preProcessPattern';
 import processPattern from './processPattern';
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -46,6 +47,15 @@ function CopyWebpackPlugin(patterns = [], options = {}) {
         let contextDependencies;
         const written = {};
 
+        let context;
```
michael-ciniawsky: `\n` between the variable declaration and the if block please :)

cletusw: Should I place it up with the other `let`s and `const`s as well?

michael-ciniawsky:
@cletusw Please rebase when time :)

cletusw:
@michael-ciniawsky Awesome! Let me know when these patches get a version bump, thanks!

PR #152, opened 2017-10-06
chore(package): update `loader-utils` v0.2.15...1.1.0

### LINKED ISSUE #137: [Bug] DOT files are always treated as directories
DOT files are always treated as directories.

{from: path.resolve(__dirname, 'application/admin/.htaccess'), to: 'admin/.htaccess', toType: 'file'}

will result in a folder /admin/.htaccess/ being created when a file /admin/.htaccess should be created. If the "to" is changed to something like 'admin/zzz.htaccess' then it creates it as a file. It appears that the "toType" is ignored if it has a leading dot.

evilebottnawi:
### `Noteable Changes`

- Some code was removed because he is fixed problems in old `loader-utils`.
- Use `absolute` path and `context` to right determinate interpolate name.

### `Issues`

- Fixes #137

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "find-cache-dir": "^1.0.0",
     "serialize-javascript": "^1.4.0",
     "is-glob": "^4.0.0",
-    "loader-utils": "^0.2.15",
+    "loader-utils": "^1.1.0",
     "minimatch": "^3.0.4",
     "p-limit": "^1.0.0"
   },
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -67,37 +67,20 @@ export default function writeFile(globalRef, pattern, file) {
             if (pattern.toType === 'template') {
                 info(`interpolating template '${file.webpackTo}' for '${file.relativeFrom}'`);
 
-                // A hack so .dotted files don't get parsed as extensions
-                let basename = path.basename(file.relativeFrom);
-                let dotRemoved = false;
-                if (basename[0] === '.') {
-                    dotRemoved = true;
-                    file.relativeFrom = path.join(path.dirname(file.relativeFrom), basename.slice(1));
-                }
-
                 // If it doesn't have an extension, remove it from the pattern
                 // ie. [name].[ext] or [name][ext] both become [name]
                 if (!path.extname(file.relativeFrom)) {
                     file.webpackTo = file.webpackTo.replace(/\.?\[ext\]/g, '');
                 }
 
-                // A hack because loaderUtils.interpolateName doesn't
-                // find the right path if no directory is defined
-                // ie. [path] applied to 'file.txt' would return 'file'
-                if (file.relativeFrom.indexOf(path.sep) < 0) {
-                    file.relativeFrom = path.sep + file.relativeFrom;
-                }
-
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "fs-extra": "^4.0.2",
     "glob": "^7.1.2",
     "is-glob": "^4.0.0",
-    "loader-utils": "^0.2.15",
+    "loader-utils": "^1.1.0",
```
michael-ciniawsky: This is strictly a 🏷  Major 😞 . We cpuld do it, if we find consensus about being a bit looser with semver here, since node =< 4 is EOL anyways 😛

src/writeFile.js
```diff
@@ -31,37 +31,20 @@ export default function writeFile(globalRef, pattern, file) {
             if (pattern.toType === 'template') {
```
michael-ciniawsky: Please split into separate PR, bc of the comment above

package.json
evilebottnawi: @michael-ciniawsky i think it is more fix then breaking changes because some interpolation don't works as expected

src/writeFile.js
evilebottnawi: @michael-ciniawsky after upgrade this code do error, need remove this lines, that everything worked as it should

package.json
evilebottnawi: And yes all that < node 4 is EOL :smile:

src/writeFile.js
michael-ciniawsky: So the dependency upgrade is required for these changes ? What exactly errors atm (I'm not too familiar with this package yet tbh 😅 )

package.json
michael-ciniawsky: I'm basically ok with it, but let's wait a bit what others say

src/writeFile.js
evilebottnawi: @michael-ciniawsky yes with .dot files and files without directory `foo.txt`

michael-ciniawsky: #137 ? If see that one, but the issue originates from `loaderUtils.interpolateName()` ?

evilebottnawi: @michael-ciniawsky this problems avoid after upgrade

d3viant0ne:
I've updated the target branch to `next` & we can land this along with #146. For simplicities sake we can land this one first so you don't have to deal with the defaults rebase & i'll pick up the changes over in #146

package.json
d3viant0ne: How about we skip the grey area on the breaking nature of loader-utils & just roll it in with defaults which is most definitely a major. We can publish to `@beta / @next / @purple_fluffie_bunnies` off the `#next` branch until everyone is satisfied with it. Also gives me a chance to test against angular/cli without blocking other PRs in master

michael-ciniawsky:
@evilebottnawi Could you rebase please, I set the `engines` field in `package.json` to `node >= v4.0.0` a while ago and there weren't any complaints so far. `node =< v4.0.0` is EOL for over a year now I think we can land this as a patch in this case :) ?

evilebottnawi:
@michael-ciniawsky I hope we will not break anything :smile:

PR #153, opened 2017-10-06
ci(travis): add node versions

deiga:
Why:

* For better coverage of stability

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,7 @@ language: node_js
 node_js:
   - "4"
   - "6"
+  - "7"
+  - "8"
+  - "node"
+  - "lts/*"


evilebottnawi:
@deiga Thanks but we do this in https://github.com/webpack-contrib/copy-webpack-plugin/pull/146, it is common repo for all webpack-contrib repos, if you wan't to improve something please write issue or do pr in `webpack-default`, thank you!

michael-ciniawsky:
Thx, but as @evilebottnawi stated why take care of these things via an org wide boilerplate :)

deiga:
Right, sorry about that! I didn't think to look at stuff after the owner change here :)

PR #164, opened 2017-11-09
feat: pass stats to transform function

axin:


axin:
@evilebottnawi I'm trying to move legacy project build from gulp to webpack. This project has file bootstrap.js which loads other js files. Code looks like this:
```js
    var src = 'js/vendors.js';
    if (window.attachEvent && !window.addEventListener) {
        window.isIe8 = true;
        src = 'js/vendors-ie8.js';
    }

    var vendorsScript = document.createElement('script');
    vendorsScript.src = src;
    vendorsScript.async = false;
    document.body.appendChild(vendorsScript);
```

vendors.js and vendors-ie8.js are hashed (looks like vendor-45627598754.js), so I cannot hardcode their names in bootstrap.js. I'm going to put placeholders, instead of file names in bootstrap.js, and then replace those placeholders with actual file names in transform function, getting actual file names from stats.

axin:
There is another option: we can pass [manifest](https://github.com/danethurber/webpack-manifest-plugin) (if manifest pluguin used) insead of whole stats.

michael-ciniawsky:
Closing in favor of #198

PR #174, opened 2017-11-23
refactor: use `.add` instead of `.push` for compilation dependencies

filipesilva:
Webpack 4 uses `SortableSet`, which doesn't have `.push`.

This PR is only for Webpack 4.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -118,7 +118,11 @@ function CopyWebpackPlugin(patterns = [], options = {}) {
                     debug(`not adding ${file} to change tracking, because it's already tracked`);
                 } else {
                     debug(`adding ${file} to change tracking`);
-                    compilation.fileDependencies.push(file);
+                    const { fileDependencies } = compilation;
+
+                    Array.isArray(fileDependencies)
+                        ? fileDependencies.push(file)
+                        : fileDependencies.add(file);
                 }
             });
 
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
@filipesilva There isn't any webpack API available to handle this I assume ? If this should be a patch instead we need to add a type check and handle it properly e.g

```js
const { contextDependencies } = compilation

if (Array.isArray(contextDependendencies)) {
   return contextDependencies.push()
}

return contextDependencies.add()
```

filipesilva:
I didn't see any new API in Webpack 4 for this. I suppose that at the end of the day it's up to you guys if you want to make `copy-webpack-plugin` support Webpack 3 and 4 at the same time or not.

If you do, then I agree that the type check is preferrable.

filipesilva:
Added the type check for now. No rush on this PR on my side btw, there might be other changes needed to support Webpack 4.

src/index.js
```diff
@@ -128,7 +132,11 @@ function CopyWebpackPlugin(patterns = [], options = {}) {
                     debug(`not adding ${context} to change tracking, because it's already tracked`);
                 } else {
                     debug(`adding ${context} to change tracking`);
-                    compilation.contextDependencies.push(context);
+                    const { contextDependencies } = compilation;
+
+                    Array.isArray(contextDependencies)
+                      ? contextDependencies.push(file)
```
michael-ciniawsky: `file` => `context`

filipesilva: Urf, fixing.

filipesilva:
Sec, messed up branched.

michael-ciniawsky:
If this isn't urgent and we don't know if other changes are required should we left this PR open until someone actually tested it with `webpack@next` ?

filipesilva:
Yeah I think it's better to not merge yet.

filipesilva:
This doesn't seem to be needed with `4.0.0-alpha.0`, closing.

PR #176, opened 2017-11-24
feat: add option to cache `pattern.transform` (`pattern.cache`)

evilebottnawi:
### `Notable Changes`

- [x] implement `serialize` for cache key to invalidate cache after `pattern.transform` changed
- [x] add tests

### DIFF
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -2,6 +2,10 @@ import fs from 'fs';
 import pify from 'pify';
 import loaderUtils from 'loader-utils';
 import path from 'path';
+import cacache from 'cacache';
+import serialize from 'serialize-javascript';
+import { name, version } from '../package.json';
+import findCacheDir from 'find-cache-dir';
 
 export default function writeFile(globalRef, pattern, file) {
     const {info, debug, compilation, fileDependencies, written, copyUnmodified} = globalRef;
@@ -22,7 +26,42 @@ export default function writeFile(globalRef, pattern, file) {
         return pify(fs.readFile)(file.absoluteFrom)
         .then((content) => {
             if (pattern.transform) {
-                content = pattern.transform(content, file.absoluteFrom);
+                const transform = (content, absoluteFrom) => {
+                    return pattern.transform(content, absoluteFrom);
+                };
+
+                if (pattern.cache) {
+                    if (!globalRef.cacheDir) {
+                        globalRef.cacheDir = findCacheDir({ name: 'copy-webpack-plugin' });
+                    }
+
+                    const cacheKey = pattern.cache.key
+                        ? pattern.cache.key
+                        : serialize({
+                            name,
+                            version,
+                            pattern,
+                            content
+                        });
+
+                    return cacache
+                    .get(globalRef.cacheDir, cacheKey)
+                    .then(
+                         (result) => JSON.parse(result.data),
+                          () => {
+                              return Promise
+                              .resolve()
+                              .then(() => transform(content, file.absoluteFrom))
+                              .then((content) => cacache.put(
+                                  globalRef.cacheDir,
+                                  cacheKey,
+                                  JSON.stringify(content)
+                               ).then(() => content));
+                          }
+                     );
+                }
+
+                content = transform(content, file.absoluteFrom);
             }
 
             return content;
--- a/tests/index.js
+++ b/tests/index.js
@@ -9,6 +9,8 @@ const CopyWebpackPlugin = require('./../dist/index');
 import fs from 'fs';
 import path from 'path';
 import _ from 'lodash';
+import findCacheDir from 'find-cache-dir';
+import cacache from 'cacache';
 
 const BUILD_DIR = path.join(__dirname, 'build');
 const HELPER_DIR = path.join(__dirname, 'helpers');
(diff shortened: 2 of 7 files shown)


src/writeFile.js
```diff
@@ -23,7 +26,34 @@ export default function writeFile(globalRef, pattern, file) {
         return fs.readFileAsync(file.absoluteFrom)
         .then((content) => {
             if (pattern.transform) {
-                content = pattern.transform(content, file.absoluteFrom);
+                const transform = (content, absoluteFrom) => {
+                    return pattern.transform(content, absoluteFrom);
+                };
+
+                if (pattern.cache) {
+                    const cacheKey = pattern.cacheKey
```
michael-ciniawsky: Do we really need to ability to customize  the `cacheKey` via `pattern.options` ?

src/writeFile.js
```diff
@@ -23,7 +26,34 @@ export default function writeFile(globalRef, pattern, file) {
...
+                        ? pattern.cacheKey
+                        : JSON.stringify(serialize({
```
michael-ciniawsky: Could we use/generate an `id` or `hash`  for patterns to avoid serialization ?

src/writeFile.js
evilebottnawi: @michael-ciniawsky yep, we can use package inside `transform` function which get options, to invalidate cache you should pass this variables as cacheKey

src/writeFile.js
evilebottnawi: @michael-ciniawsky we can generate md5 or sha algo for this

evilebottnawi:
@michael-ciniawsky to allow cache between webpack builds, we do this in `uglify` plugin

src/writeFile.js
evilebottnawi: @michael-ciniawsky after testing with difference arguments and platform, seems `serialize` is best solution, btw we can do this in `uglify` plugin

evilebottnawi:
@michael-ciniawsky friendly ping

src/writeFile.js
```diff
@@ -22,7 +26,42 @@ export default function writeFile(globalRef, pattern, file) {
...
+                const transform = (content, absoluteFrom) => {
+                    return pattern.transform(content, absoluteFrom);
+                };
+
+                if (pattern.cache) {
+                    if (!globalRef.cacheDir) {
+                        globalRef.cacheDir = findCacheDir({ name: 'copy-webpack-plugin' });
+                    }
+
+                    const cacheKey = pattern.cacheKey
```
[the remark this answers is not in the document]
evilebottnawi: @michael-ciniawsky we need another opinion :smile:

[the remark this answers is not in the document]
evilebottnawi: @michael-ciniawsky btw we can remove this option, because we use `pattern` as part of cache key and if you want invalidate cache you can use `cache` option (example `cache: '/path/to/' + crypto.createHash('md5').update(JSON.stringify(something) || serialize(something)).digest("hex")`)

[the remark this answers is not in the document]
michael-ciniawsky: @evilebottnawi If you add it to the `cache` pattern option itself `{ cache: { key: '...' } } || { cache: key }` I'm ok with it, but adding to many (related) options 'top scope' of a pattern is unnecessary (bloat) and can be confusing for folks new to `copy-webpack-plugin` e.g if they assume `pattern.cacheKey` is mandatory to use for some reason (many don't fully read the docs). There are already too many options for patterns and most folks will tend to use simple patterns like `{ from: '', to: '' } || { from: '', to: '', cache: true }` which is sufficient && straight forward for most of the plugin use cases.

tests/index.js
```diff
@@ -1305,5 +1308,181 @@ describe('apply function', () => {
...
+                })
+                .then(() => {
+                    return cacache
+                        .ls(cacheDir)
+                        .then((cacheEntries) => {
+                            const cacheKeys = Object.keys(cacheEntries);
+
+                            expect(cacheKeys).to.have.lengthOf(1);
+
+                            cacheKeys.forEach((serializedCacheEntry) => {
```
michael-ciniawsky:
```js
cacheKeys.forEach((cacheKey) => {
  // eslint-disable-next-line no-eval
  cacheKey = eval(`(${cacheKey})`);
 
  expect(cacheKey.pattern.from).to.equal(from);
});

src/writeFile.js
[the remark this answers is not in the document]
evilebottnawi: @michael-ciniawsky btw, sometimes i have external variables for function inside in `transofrm` and i can't invalidate cache without this option, i can only enable or disable cache, but this not convenient. Example i use `nunjucks` inside `transform`, some template render long time and i want to cache this operation, also i know when i should invalidate cache and cacheKey solve this problem.

tests/index.js
evilebottnawi: @michael-ciniawsky done

src/writeFile.js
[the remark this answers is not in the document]
michael-ciniawsky: Yep, but this already goes under 'advanced' usage in the scope of `copy-webpack-plugin` imo and wouldn't this work with `{ cache: ${customCacheKey}` instead of `{ cache: true, cacheKey: `${customCacheKey}` }` for some reason ?

evilebottnawi:
@michael-ciniawsky friendly ping, move key inside `cache` option.

PR #177, opened 2017-11-24
docs(README): replace `var` with `const` in examples

capaj:
promote best practice in the docs


evilebottnawi:
@capaj please accept CLA

capaj:
just did

PR #178, opened 2017-11-24
refactor: use native `{Promise}` instead of `bluebird`

evilebottnawi:
Node which we support have native implementation `Promise`.

### DIFF
--- a/src/processPattern.js
+++ b/src/processPattern.js
@@ -1,10 +1,9 @@
-import Promise from 'bluebird';
+import globby from 'globby';
+import pLimit from 'p-limit';
 import path from 'path';
 import _ from 'lodash';
 import minimatch from 'minimatch';
 import writeFile from './writeFile';
-
-const globAsync = Promise.promisify(require('glob')); // eslint-disable-line import/no-commonjs
 
 export default function processPattern(globalRef, pattern) {
     const {info, debug, output, concurrency} = globalRef;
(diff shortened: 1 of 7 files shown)


src/processPattern.js
```diff
@@ -1,11 +1,10 @@
-import Promise from 'bluebird';
+import globby from 'globby';
+import createThrottle from 'async-throttle';
```
michael-ciniawsky: [p-all](https://github.com/sindresorhus/p-all)
[p-limit](https://github.com/sindresorhus/p-limit)
[p-throttle](https://github.com/sindresorhus/p-throttle)

[async-throttle](https://github.com/zeit/async-throttle) is **deprecated**

src/processPattern.js
```diff
@@ -16,9 +15,11 @@ export default function processPattern(globalRef, pattern) {
         return Promise.resolve();
     }
 
+    const throttle = createThrottle(concurrency || 100);
+
     info(`begin globbing '${pattern.absoluteFrom}' with a context of '${pattern.context}'`);
-    return globAsync(pattern.absoluteFrom, globArgs)
-    .map((fileFrom) => {
+    return globby(pattern.absoluteFrom, globArgs)
+    .then((filePaths) => Promise.all(filePaths.map((fileFrom) => throttle(() => {
```
michael-ciniawsky: Please indent 2 by spaces, `filePaths` => `files || paths`, `fileFrom` => `file || from`

src/processPattern.js
evilebottnawi: @michael-ciniawsky archived, not deprecated :+1:

d3viant0ne: None the less I agree with @michael-ciniawsky , i'd prefer using something that is actively maintained

evilebottnawi:
@michael-ciniawsky @d3viant0ne friendly ping

PR #179, opened 2017-11-24
feat: add `write` pattern option (`pattern.write`)

evilebottnawi:


evilebottnawi:
@d3viant0ne @michael-ciniawsky friendly ping

michael-ciniawsky:
@evilebottnawi I'm not really sure about this feature tbh, it feels redundant to some point I need to think about it first

evilebottnawi:
@michael-ciniawsky this need for server side rendering application (php, python, ruby and etc), example popular plugin for manifest have this option https://github.com/danethurber/webpack-manifest-plugin#optionswritetofileemit

README.md
```diff
@@ -42,6 +42,7 @@ Or, in the simple case of just a `from` with the default destination, you can us
 | `ignore` | N | [] | Additional globs to ignore for this pattern |
 | `transform` | N | function(content, path) {<br>&nbsp;&nbsp;return content;<br>} | Function that modifies file contents before writing to webpack |
 | `force` | N | false | Overwrites files already in compilation.assets (usually added by other plugins) |
+| `writeWhenEmit` | N | false | If set to `true` will emit to build folder and memory in combination with `webpack-dev-server` |
```
michael-ciniawsky: Can we find a terser name for this `write || emit`?

src/writeFile.js
```diff
@@ -1,4 +1,4 @@
-import fs from 'fs';
+import fs from 'fs-extra';
```
michael-ciniawsky: `mkdirp` instead of the 'whole' `fs-extra` dependency?

tests/index.js
```diff
@@ -6,9 +6,13 @@ import {
 // ensure we don't mess up classic imports
 const CopyWebpackPlugin = require('./../dist/index');
 
-import fs from 'fs';
+import fs from 'fs-extra';
 import path from 'path';
 import _ from 'lodash';
+import chai from 'chai';
```
michael-ciniawsky: Please no `chai` => [`jest.mock`]( http://facebook.github.io/jest/docs/en/mock-function-api.html#content) 🙃

michael-ciniawsky:
https://www.npmjs.com/package/write-file-webpack-plugin

README.md
evilebottnawi: @michael-ciniawsky yep, write seems good for me

src/writeFile.js
evilebottnawi: @michael-ciniawsky `mkdirp` have many issue with windows, `fs-extra` more stable

tests/index.js
evilebottnawi: @michael-ciniawsky i think we do this when merge with default

src/writeFile.js
michael-ciniawsky: kk

tests/index.js
michael-ciniawsky: hmm... 😞  fair point but it's still needs to be refactored again with `webpack-defaults` then 🤢. Can't be helped

evilebottnawi:
@michael-ciniawsky seems good plugin, thanks! Let's wait other opinion about this feature

michael-ciniawsky:
If that plugin is sufficient enough (e.g not sure if ignoring some files is possible with it) we should avoid adding more `pattern` options

evilebottnawi:
@michael-ciniawsky i test this plugin in near future, if that plugin is sufficient enough, i will close this PR

evilebottnawi:
Close in favor write-file-webpack-plugin

PR #187, opened 2017-12-14
chore(package): add `files` property to reduce package size

evilebottnawi:
:star:

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,0 @@
-src
-tests
-compiled_tests
-coverage
-.*
-*.log
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "author": "Len Boyette",
   "license": "MIT",
   "main": "dist/index.js",
+  "files": ["dist"],
   "scripts": {
     "lint": "eslint src/ tests/",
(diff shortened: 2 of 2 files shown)


michael-ciniawsky:
Use `pkg.files` instead 🙃 ?

evilebottnawi:
@michael-ciniawsky just for this repo (before `webpack-default`), because we have .npmignore :smile:

michael-ciniawsky:
Please use `pkg.files` as it is the cleaner solution for this and reduces the overall package size as a bonus, regardless of `webpack-defaults` or not 🙏

evilebottnawi:
@michael-ciniawsky friendly ping

michael-ciniawsky:
> In the short term, just use `files` (pkg.files)

🙃 You can leave `appveyor.yml` in `.npmignore` if you want, but `files` simply the better solution here

evilebottnawi:
@michael-ciniawsky we migrate to files after `webpack-default`, just little fix :smile: We spend many time on this simple issue :laughing:

evilebottnawi:
@michael-ciniawsky @d3viant0ne rewrite using `files`

PR #191, opened 2017-12-15
refactor: avoid using `lodash`

evilebottnawi:


src/preProcessPattern.js
```diff
@@ -13,7 +12,7 @@ export default function preProcessPattern(globalRef, pattern) {
 
     pattern = typeof pattern === 'string' ? {
         from: pattern
-    } : _.cloneDeep(pattern);
+    } : Object.assign({}, pattern);
```
michael-ciniawsky: This might be ok here, but `Object.assign` doesn't deep clone `{Object}`'s to my knowledge

src/preProcessPattern.js
```diff
@@ -39,9 +38,13 @@ export default function preProcessPattern(globalRef, pattern) {
     debug(`determined '${pattern.to}' is a '${pattern.toType}'`);
 
     // If we know it's a glob, then bail early
-    if (_.isObject(pattern.from) && pattern.from.glob) {
+    if (Object.prototype.toString.call(pattern.from) === '[object Object]' && pattern.from.glob) {
```
michael-ciniawsky: `typeof pattern.from === 'object'` ?

src/preProcessPattern.js
d3viant0ne: That's not going to work. You will need to add a lib that actually does a deep object clone.

d3viant0ne:
What exactly is the issue with using lodash?

It's a lib that's never going to be published to the web & even if it was, uglify-es is one of the most tree-shakable libs in existance

michael-ciniawsky:
Anything that has a standard ES2015+ equivalent can/should be dropped imho but e.g for deep cloning we need/should stick with `lodash` nothing wrong about that :)

src/preProcessPattern.js
evilebottnawi: @d3viant0ne This line doesn't have sense with merge deep

src/preProcessPattern.js
evilebottnawi: @michael-ciniawsky `typeof null === 'object'` and `typeof [1,2] === 'object'`
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

evilebottnawi:
Exactly, native function more faster

michael-ciniawsky:
@evilebottnawi What's the status here ? If ready for review/merge could you rebase please ? :)

evilebottnawi:
@michael-ciniawsky Today I will finish :+1:

evilebottnawi:
@michael-ciniawsky done :+1:

src/preProcessPattern.js
michael-ciniawsky: Could you move this to into a helper function `isObject` instead please ? 

```js
const isObject = (val) => Object.prototype.toString.call(val) === '[object Object]' ? true : false
```

michael-ciniawsky:
Just one small nit :)

evilebottnawi:
@michael-ciniawsky done :+1:

Issue #105, opened 2016-12-21
Manifest for copied assets


liqueflies:
Hi,
what about manifest for hash copied assets?
There are some plugins that work with copy-webpack?

Thank you.

pinkstrings:
I found a hacky way around this using the `transform` option and the `md5-file` module:

```javascript
new CopyWebpackPlugin([
  {
    from: 'foo.zip',
    to: 'foo-[hash].zip',
    transform: function(content, filepath) {
      var manifestContents = {'foo.zip': 'foo-' + require('md5-file').sync(filepath) + '.zip'}
      require('fs').writeFileSync('manifest.json', JSON.stringify(manifestContents))
      return content
    }
  }
])
```

imerr:
I'm running into the same issue here trying to use laravel mix (which is a wrapper around webpack to simplify common things)
They use webpack-stats-plugin to write out a manifest file
Which seems to do something like this https://github.com/FormidableLabs/webpack-stats-plugin/issues/3 :
```js
    function() {
      this.plugin('done', function(stats) {
        require('fs').writeFileSync(
          './dist/manifest.json',
          JSON.stringify(stats.toJson().assetsByChunkName));
      });
    }
```

I can't for the live of me figure out what the critera for files to show up in stats is
Could someone more familiar with webpack maybe give me some pointers?

SebastianS90:
If you know the hash length, then you can use `StatsWriterPlugin` and extract hashes from the `stats.asset.name` field as presented in https://github.com/kevlened/copy-webpack-plugin/issues/104#issuecomment-280915911

davidmpaz:
@SebastianS90 how can I make it work with directories. Let say I am copying like this:
```
config.plugins.push(new CopyWebpackPlugin([
    {
        from: resourcesPath + '/images',
        to: outputPath
    }
]));
```

Curently I dont see a way to overcome with this. 

It would be very nice if this is added to this plugin.

SebastianS90:
@davidmpaz I also don't see an easy way to get a manifest for a mapping from source file to output file.
You can have a look at the [example stats object](https://raw.githubusercontent.com/webpack/analyse/master/app/pages/upload/example.json) and there you see the `assets` property. It does not contain the source of each asset.

What my solution does is to take the output file path, remove the hash and create a mapping from the unhashed to the hashed output file path. For my use-case, I reference files via their unhashed output file and use the created manifest to find the hashed output file.

[...]

Issue #86, opened 2016-11-04
[Feature] Event Hooks


QWp6t:
In my particular use case, I'm wanting to take the initial file name and output file name (in my case, hashed), and then use those to write to a manifests file. Having events in this plugin into which I could hook to get the file names would be awesome. There might be other use cases for events in other places as well, which is why I wasn't super specific with my issue title.

So if nothing else, would be possible to [fire an event right about here](https://github.com/kevlened/copy-webpack-plugin/blob/323b1d74ef35ed2221637d8028b1bef854deb523/src/writeFile.js#L83) with perhaps a signature of `function(file){}`?

Then I could use properties on `file` (`relativeFrom` and `webpackTo`, to be specific) to write my manifest.

QWp6t:
I can submit a PR with the event I suggested, if you'd be receptive to it.

kevlened:
I wonder if there's a way to integrate with [webpack-manifest-plugin](https://github.com/danethurber/webpack-manifest-plugin) or [webpack-assets-manifest](https://www.npmjs.com/package/webpack-assets-manifest) instead of adding an extra event. This needs more investigation.

QWp6t:
Yes, that's definitely possible. I do it in a [similar plugin that I wrote that's based on yours](https://github.com/roots/sage/blob/alpha4/assets/build/webpack.plugin.copyglobs.js#L72) (I assign the value to the manifest on [L94](https://github.com/roots/sage/blob/alpha4/assets/build/webpack.plugin.copyglobs.js#L94). I made a point to use the same property names as you so that it would be easy to read. As you can see, the manifest object is populated like `{ relativeFrom: webpackTo }`

[How it's used](https://github.com/roots/sage/blob/alpha4/assets/build/webpack.config.js#L122-L128):

```js
const myAwesomeManifest = {};
// ...
  plugins:
    new CopyGlobsPlugin({
      // It would be nice to switch to copy-webpack-plugin, but unfortunately it doesn't
      // provide a reliable way of tracking the before/after file names
      pattern: 'images/**/*',
      output: `[path][name]_[hash:8].[ext]`,
      manifest: myAwesomeManifest,
    }),
    new WebpackAssetsManifest({
      output: 'assets.json',
      assets: config.manifest,
    }),
  ]
```

michael-ciniawsky:
Whats is the status here ? Still any need/interest in resolving this ? :)

Levdbas:
[...]

PR #198, opened 2017-12-26
feat: add `manifest` support (`options.manifest`)

### LINKED ISSUE #104: [Feature] Support manifest file
Why?
In some projects, I use external library (example `jquery`) and for the implamentation of this, I use this plugin (just copy files from another directory to other). The plugin supports the use of `hash`, but do not know original file path and resulting file path (include `hash` and other stuff). There is no manifest file :sob:

QWp6t:
This seems to be a popular feature request so I decided to take a stab at it.

I had restore some of the modifications that get made to `file.relativeFrom` in src/writeFile.js. Since `file.relativeFrom` isn't used after that point, it shouldn't affect the outcome.

Users have the option of passing a function instead, which is probably a more useful option as it will allow them to control the structure of their manifest object.

### `Issues`

- Fixes #104


QWp6t:
x-ref: #86, #104, #105.

tests/index.js
```diff
@@ -1573,5 +1573,61 @@ describe('apply function', () => {
...
+                    'directoryfile.txt': 'dist/directoryfile-22af64.txt',
+                    'nested/nestedfile.txt': 'dist/nested/nestedfile-d41d8c.txt'
+                };
+                
+                run({
+                    patterns: [{
+                        from: 'directory',
+                        to: 'dist/[path][name]-[hash:6].[ext]'
+                    }],
+                    options: { manifest: myManifest }
```
michael-ciniawsky: `options: { manifest: myManifest }` => `options: { manifest }`

src/writeFile.js
```diff
@@ -71,6 +71,7 @@ export default function writeFile(globalRef, pattern, file) {
                 // A hack so .dotted files don't get parsed as extensions
                 let basename = path.basename(file.relativeFrom);
                 let dotRemoved = false;
+                let slashAdded = false;
```
michael-ciniawsky: Can you elaborate why this is needed ?

QWp6t: Because the slash is removed under certain circumstances. It's documented within the source of that file. (I'm on my phone right now. I can explain it in more detail later if necessary, but I think you can find the reason already documented in the file.)

QWp6t: Sorry, I said "the slash is removed," but I meant a leading slash is *added*. Anyway,  yeah, the reason for adding the leading slash is [explained within the file](https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/writeFile.js#L85-L90). The reason I remove it again is to ensure that the manifest keys are consistent.

Honestly, I'd say that `file.relativeFrom` should go completely unaltered in that `(pattern.toType === 'template')` block, and a new variable should instead be created and passed to `loaderUtils.interpolateName()`. I didn't do that here since that would be a refactor, which I think would be outside the scope of this PR.

michael-ciniawsky: All good, that's obviously on our side then :). Could you please add a comment with the link `https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/writeFile.js#L85-L90` as an explanation/context for why this is needed atm?

tests/index.js
```diff
@@ -1573,5 +1573,61 @@ describe('apply function', () => {
...
+                })
+                .then(() => {
+                    expect(manifest).to.deep.equal(expectedManifest);
+                })
+                .then(done)
+                .catch(done);
+            });
+
+            it('should accept a manifest function', (done) => {
+                const myManifest = {
```
michael-ciniawsky: Also just `manifest` here please

src/writeFile.js
```diff
@@ -85,7 +85,9 @@ export default function writeFile(globalRef, pattern, file) {
                 // A hack because loaderUtils.interpolateName doesn't
                 // find the right path if no directory is defined
                 // ie. [path] applied to 'file.txt' would return 'file'
+                let slashAdded = false;
```
QWp6t: I moved it down here so it's consistent with the `dotRemoved` block. So now your guys' code comments also serve as comments for the existence of the newly introduced variable. Hope this will suffice.

deguif:
Any status on this?
Would really like to be able to use it.

QWp6t:
Note to devs: this PR will have merge conflicts with https://github.com/webpack-contrib/copy-webpack-plugin/pull/152, but that one already has merge conflicts so I'd suggest merging this one first. I can help with conflict resolution on the other one after this one is merged if you'd like.

evilebottnawi:
@michael-ciniawsky @d3viant0ne seems good, but i think best solution add use event on this as in https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L180

QWp6t:
@evilebottnawi that's exactly what my original suggestion was. ( see: #86 ) ;)

But this PR does also work similarly. An object or a function can be accepted. The README.md contains the function signature. Basically, I'm trying to please everyone with this PR, but as I noted in the first post, I think that the function is a much more useful option (regardless of whether it's an event handler or a callback).

Baedda:
Please merge. We really need this.

evilebottnawi:
@michael-ciniawsky what do you think about this?

jaredwy:
It bothers me that you have to write your own file either way to use this plugin. I either pull the manifest from some other plugin, e.g. the webpack-manifest-plugin then pass it into this function and then wait until webpack is complete and write out my file. 

Or i use the  the file call back either means we have to buffer the object or open/close a file on every file copy for write on every file copied. 

There is also no assurance that webpack-manifest-plugin (or others) won't change in future versions to make this harder either i.e. if they update to subscribe to only write the file on `after-emit`? 

Please correct me if I have missed something obvious, I am just finding it hard to use this addition.

DASPRiD:
Really looking forward to this feature, as this is currently blocking me.

@QWp6t BTW: you should **never** merge master into your branch, but rather rebase your branch against master.

laozhu:
Any status on this?

tomm1996:
maybe release a beta or some branch with this PR merged so people can use this essential feature pls?

src/writeFile.js
michael-ciniawsky: `slashAdded` => `hasSlash`

michael-ciniawsky:
@evilebottnawi Could you review again if something needs to be changed here since a few things landed in on master in the meantime? Otherwise +1 from my side for the feature :)

michael-ciniawsky:
@QWp6t Please rebase :)

evilebottnawi:
@michael-ciniawsky better add event inside plugin for this as do html-webpack-plugin and do it using hooks :+1:

deguif:
Hi,

After waiting a long time for this PR to be merged, I finally found a solution.
For anyone wanting to use [danethurber/webpack-manifest-plugin](https://github.com/danethurber/webpack-manifest-plugin) and be able to generate asset manifest which were copied with this plugin, here's my solution:

In webpack.config.js
```js
const path = require('path');
const CopyPlugin = require('copy-webpack-plugin');
const ManifestPlugin = require('webpack-manifest-plugin');
...

module.exports = function (env) {
    return {
        ...
        output: {
            path: path.resolve(__dirname, 'web/dist'),
            publicPath: '/dist/',
        },
        plugins: [
            new CopyPlugin([
                {
                     context: path.join('/xxx/yyy', 'images'),
                     from: '**/*{.gif,.jpg,.jpeg,.png,.svg}',
                     to: 'images/[path][name].[ext]',
                     // to: 'production' === env.NODE_ENV ? 'images/[path][name].[hash].[ext]' : 'images/[path][name].[ext]',
                     toType: 'template',
                },
            ]),
            // Order matters, manifest plugin should be registered after copy plugin
            new ManifestPlugin({
                fileName: path.resolve(__dirname, 'var/assets/images/manifest.json'),
                publicPath: '/dist/',
                filter: (file) => file.name.startsWith('images/'),
                /*map: (file) => {
                    if ('production' === env.NODE_ENV) {
                        // Remove hash in manifest key
                        file.name = file.name.replace(/(\.[a-f0-9]{32})(\..*)$/, '$2');
                    }
                    return file;
                },*/
            }),
        ],
    };
}
```

__This will only work with version 2.x of [danethurber/webpack-manifest-plugin](https://github.com/danethurber/webpack-manifest-plugin) currently in rc2__
See issue danethurber/webpack-manifest-plugin#75 and commit danethurber/webpack-manifest-plugin@4cb95ce41ec683078a6581949ec544f4a7d4a3d2

My example copy images from a `/xxx/yyy/images` to a `web/dist/images` folder and generate a manifest for images only in `var/assets/images/manifest.json` ([Symfony 3](https://symfony.com) project). For using hash when copying the assets you will need a little adaption on manifest plugin, you can use the `map`hook to remove hash from filename (See my commented code).

I hope this will help people from this thread.

QWp6t:
Closing this in favor of https://github.com/webpack-contrib/copy-webpack-plugin/pull/309

PR #200, opened 2017-12-28
add chunk name for identification

lcxfs1991:
add a chunk attribute for other plugins to identify the resource

### DIFF
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -7,6 +7,23 @@ import serialize from 'serialize-javascript';
 import { name, version } from '../package.json';
 import findCacheDir from 'find-cache-dir';
 
+function getChunkName(resourcePath, fileWepackTo) {
+    var i;
+
+    for (i = fileWepackTo.length - 1; i >= 0; i--) {
+        if (fileWepackTo[i] === '/' || fileWepackTo[i] === '\\') {
+            break;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@lcxfs1991 can you describe use case?

lcxfs1991:
For many plugins like manifest or html plugin, they usually need to locate asset copied by `copy-webpack-plugin`. However, `stats` object of webpack is not user friendly, so I create another shortcut way to help developers to locate these resources.

evilebottnawi:
@lcxfs1991 stats very friendly, what kind of problem with stats you have?

lcxfs1991:
image

For example, this is an example of stats generated by webpack. libs/preact and libs/react is copied by `copy-webpack-plugin`. The `name` is filled with the name with hash key. So, it is really hard to be identified by other plugins. So I add an attribute for other plugins to identify the original file name / key name of the asset.

evilebottnawi:
@lcxfs1991 maybe best add `originalName` to stats (Issue and PR to webpack)?

lcxfs1991:
I also think it is a good idea to pr webpack. But webpack is a big project and the life cycle maybe too long.

evilebottnawi:
@lcxfs1991 i don't think, just do PR in webpack and we can release 3.10.1, maybe some fixes also wait 3.10.1 release. I ask about this sokra in slack channel, but PR in `webpack` really good idea :+1:

Issue #204, opened 2018-01-15
Is it possible to use something like compiler.inputFileSystem over fs?


Meberem:
Would it be possible to pass the `compiler.inputFileSystem` or `compilation.inputFileSystem` or similar to the functions that need it as opposed to taking a hard dependency on the fs module?

It would make testing webpack with a memoryfs a bit easier

evilebottnawi:
@Meberem look https://github.com/webpack/webpack/issues/1562#issuecomment-165506401, you can write simple script

Meberem:
Thanks for the pointer, I have been using that method however and am encountering issues such as `[copy-webpack-plugin] WARNING - unable to locate '/mock/src/a.css' at '/mock/src/a.css'`

I have distilled my issue into the following script:
```
var path = require("path");
var webpack = require("webpack");
var MemoryFileSystem = require("memory-fs");
var CopyWebpackPlugin = require("copy-webpack-plugin");

var mockFs = new MemoryFileSystem();
mockFs.mkdirpSync("/mock");
mockFs.mkdirpSync("/mock/src");
mockFs.writeFileSync("/mock/src/a.css", "a_css");
mockFs.mkdirpSync("/mock/output");

const options = {
    context: "/mock/",
    entry: "/mock/src/entry.js",
    output: {
        path: "/mock/output/",
    },
    plugins: [
        new CopyWebpackPlugin([
            {
                from: "/mock/src/a.css",
                to: "/mock/output/a.css"
            }
        ])
    ]
};

var compiler = webpack(options);

compiler.inputFileSystem = mockFs;
compiler.resolvers.normal.fileSystem = compiler.inputFileSystem;
compiler.resolvers.context.fileSystem = compiler.inputFileSystem
compiler.outputFileSystem = mockFs;

compiler.run(function(err, stats) {
    const stat = mockFs.statSync("/mock/output/a.css"); // File doesn't exist as it was never copied
    console.log(stat);
});
```

From this it looks like the issue is occuring because `/src/writeFile.js` and `/src/preProcessPattern.js` uses `fs` over a webpack provided one. Although admittedly I am fairly clueless in this area, so any pointers would be appreciated.

evilebottnawi:
@Meberem seems you are right, we should use `compiler.inputFileSystem` instead `fs` module. It try to send PR today.

Meberem:
Amazing, thank you so much

PR #205, opened 2018-01-16
feat: use `compiler.inputFileSystem` instead `fs`

evilebottnawi:
### `Notable Changes`

Also it is improve perf, because `webpack` cache `stats` and other stuff (https://github.com/webpack/webpack/blob/efc576c8b744e7a015ab26f1f46932ba3ca7d4f1/lib/node/NodeEnvironmentPlugin.js#L14) using `CachedInputFileSystem`

### `Issues`

- Closes https://github.com/webpack-contrib/copy-webpack-plugin/issues/204

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "babel-preset-es2015": "^6.6.0",
     "chai": "^3.4.0",
     "eslint": "^2.9.0",
+    "enhanced-resolve": "^3.4.1",
     "mocha": "^2.4.5",
     "ncp": "^2.0.0",
     "standard-version": "^4.2.0",
--- a/src/index.js
+++ b/src/index.js
@@ -74,6 +74,7 @@ function CopyWebpackPlugin(patterns = [], options = {}) {
                 fileDependencies,
                 contextDependencies,
                 context,
+                inputFileSystem: compiler.inputFileSystem,
                 output: compiler.options.output.path,
                 ignore: options.ignore || [],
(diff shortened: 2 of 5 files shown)


michael-ciniawsky:
@evilebottnawi Should we flag this as `perf` or `feature` instead (for the `CHANGELOG`) ?

evilebottnawi:
@michael-ciniawsky o don't know :smile: and perf and feature and refactor :smile:

michael-ciniawsky:
kk going with `feat` for now as long as nobody objects, so it definitely appears in the`CHANGELOG` 😛

michael-ciniawsky:
@evilebottnawi A rebase one more time please :) This is definitely ready right ?

package.json
```diff
@@ -31,6 +31,7 @@
     "babel-preset-es2015": "^6.6.0",
     "chai": "^3.4.0",
     "eslint": "^2.9.0",
+    "enhanced-resolve": "^3.4.1",
```
michael-ciniawsky: Is this needed as direct `devDependency` when `webpack` is already 'installed' ?

evilebottnawi: @michael-ciniawsky for tests locally

michael-ciniawsky: Sure, but `enhanced-resolve` is installed via `webpack` already right and flatten in `node_modules` ? Not important... :). Could you please rebase/generate a new lockfile when time ?

evilebottnawi:
@michael-ciniawsky done :+1:

PR #208, opened 2018-01-24
fix(preProcessPatterns): support glob context paths with special characters

clydin:
`context` paths with special characters (i.e., glob characters) will cause files to not be copied.  This is especially important when the context represents the absolute path to a location and the path is not under the control of the project using the plugin.

Tests have been updated and additional test cases added.

Addresses: https://github.com/angular/angular-cli/issues/9230


clydin:
Oddly, `apply function options ignore ignores all files except those with dots` fails on node 4/6 but passes on 8.  Will investigate.

clydin:
Alright.  Update to be more comprehensive and with additional test cases.

src/preProcessPattern.js
```diff
@@ -7,6 +7,29 @@ import isObject from './utils/isObject';
 // https://www.debuggex.com/r/VH2yS2mvJOitiyr3
 const isTemplateLike = /(\[ext\])|(\[name\])|(\[path\])|(\[folder\])|(\[emoji(:\d+)?\])|(\[(\w+:)?hash(:\w+)?(:\d+)?\])|(\[\d+\])/;
 
+function escape(context, from) {
```
evilebottnawi: Can you move `escape` to `src/utils`

clydin: Done.

PR #221, opened 2018-02-08
fix: replace `pify` with simpler promise helpers

### LINKED ISSUE #217: [Bug] after update to 4.4.0 TypeError: callbacks[i] is not a function
Compiling an angular app generated with angular cli
I have updated to 4.4.0 and I get the next error:
```
C:\Proyectos\BPS\src\webapp\node_modules\enhanced-resolve\lib\CachedInputFileSystem.js:40
                callbacks[i](err, result);
                            ^

TypeError: callbacks[i] is not a function
    at Storage.finished (C:\Proyectos\BPS\src\webapp\node_modules\enhanced-resolve\lib\CachedInputFileSystem.js:40:15)
    at C:\Proyectos\BPS\src\webapp\node_modules\enhanced-resolve\lib\CachedInputFileSystem.js:77:9
    at C:\Proyectos\BPS\src\webapp\node_modules\graceful-fs\polyfills.js:287:18
    at FSReqWrap.oncomplete (fs.js:167:5)
```

filipesilva:
`pify` seems to not be working correctly when a decorated file system is used as `inputFileSystem`.

Fix #217

### DIFF
--- a/src/utils/promisify.js
+++ b/src/utils/promisify.js
@@ -1,0 +1,21 @@
+export const stat = (inputFileSystem, path) => {
+    return new Promise((resolve, reject) => {
+        inputFileSystem.stat(path, (err, stats) => {
+            if (err) {
+                reject(err);
+            }
+            resolve(stats);
+        });
+    });
+}
+
+export const readFile = (inputFileSystem, path) => {
+    return new Promise((resolve, reject) => {
+        inputFileSystem.readFile(path, (err, stats) => {
+            if (err) {
+                reject(err);
+            }
+            resolve(stats);
+        });
+    });
+}
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "is-glob": "^4.0.0",
     "loader-utils": "^0.2.15",
     "minimatch": "^3.0.4",
-    "pify": "^3.0.0",
     "p-limit": "^1.0.0"
   },
   "devDependencies": {
--- a/src/preProcessPattern.js
+++ b/src/preProcessPattern.js
@@ -1,8 +1,8 @@
-import pify from 'pify';
 import path from 'path';
 import isGlob from 'is-glob';
 import escape from './utils/escape';
 import isObject from './utils/isObject';
+import { stat } from './utils/promisify';
 
 // https://www.debuggex.com/r/VH2yS2mvJOitiyr3
 const isTemplateLike = /(\[ext\])|(\[name\])|(\[path\])|(\[folder\])|(\[emoji(:\d+)?\])|(\[(\w+:)?hash(:\w+)?(:\d+)?\])|(\[\d+\])/;
@@ -58,7 +58,7 @@ export default function preProcessPattern(globalRef, pattern) {
 
     debug(`determined '${pattern.from}' to be read from '${pattern.absoluteFrom}'`);
 
-    return pify(inputFileSystem).stat(pattern.absoluteFrom)
+    return stat(inputFileSystem, pattern.absoluteFrom)
     .catch(() => {
         // If from doesn't appear to be a glob, then log a warning
         if (isGlob(pattern.from) || pattern.from.indexOf('*') !== -1) {
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -1,15 +1,15 @@
-import pify from 'pify';
 import loaderUtils from 'loader-utils';
 import path from 'path';
 import cacache from 'cacache';
 import serialize from 'serialize-javascript';
 import { name, version } from '../package.json';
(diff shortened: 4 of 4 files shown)


src/utils/readFilePromise.js
```diff
@@ -0,0 +1,10 @@
+export default function readFilePromise(inputFileSystem, path) {
```
michael-ciniawsky: `src/utils/promisify.js`

```js
export const stat = ...
export const readFile = ...
```

```js
import { stat, readFile } from './utils/promisify';
```

filipesilva: On it.

filipesilva: Done.

kabe5965:
New version 4.4.1 did it! Thanks!

filipesilva:
@evilebottnawi looked a bit more into why `pity` was not working with the decorated file system. It turns out `pify` uses `for (const key in obj) { }` to iterate over the object, but the decorated file system function were not enumerable and thus did not show up there.

For the decorated file system in Angular CLI we used a TypeScript class, which downlevels to a ES6 class. It seems `pify` does not work with ES6 class instances. I opened an issue for it in https://github.com/sindresorhus/pify/issues/57.

sindresorhus:
I don't understand why the fix is so overcomplicated though. You could have just used `pify` on the individual methods.

filipesilva:
@sindresorhus in hindsight, that would have been a great solution. At the time I did not know only one of the `pify` modes of operation had problems with ES6 classes.

We were looking at getting a fix in since it was affecting a lot of users in https://github.com/angular/angular-cli/issues/9550. After fixing the issue proper I went to report the issue on `pify` with a repro and found out that the other usage mode was ok.

src/utils/promisify.js
```diff
@@ -0,0 +1,21 @@
+export const stat = (inputFileSystem, path) => {
+    return new Promise((resolve, reject) => {
+        inputFileSystem.stat(path, (err, stats) => {
+            if (err) {
+                reject(err);
+            }
```
ronkorving: Without an `else` or early `return` statement, `resolve` will even be called in the case of a rejection. Same is the case for `readFile` below.

ronkorving:
I noticed what I believe to be a bug in this PR and left a comment.

src/utils/promisify.js
evilebottnawi: @ronkorving reject already do early exit, no necessary `return`

ronkorving: How can it? To my knowledge, the only way to return early in JS is `return` or `throw`.
`reject(err)` is just a function call.

evilebottnawi: @ronkorving `reject()` === `throw` inside promise

filipesilva: It is true that `resolve(stats)` will still be called, but the `stats` value will not be passed on to anything. Even if there is another promise waiting for the resolve value, the rejection path will be taken instead.

It's a still a bug though, in the sense that it was never intended for the `resolve` to run in the first place.

@evilebottnawi WDYT of me doing another PR that adds back `pify` with the alternative syntax that works on ES6 classes?

evilebottnawi: @filipesilva if you have time for this, it is not critical problem :+1:

evilebottnawi: @ronkorving IMHO, `return` here need only for easy reading and understanding code. Feel free to PR if this confusing for you

ronkorving: @evilebottnawi I don't think I'm confused. As @filipesilva also pointed out, `resolve(stats)` *will* still be called.

I'll leave it to you both to figure out the way forward, given that there seems to be a desire to put `pify` back in anyway.

filipesilva: I tried https://github.com/webpack-contrib/copy-webpack-plugin/pull/228 but just using `pify` on the single method doesn't seem to work well either, `this` seems to be undefined when doing that so class members aren't accessible.

@ronkorving would you like to make a PR to change the `if` to a `if ... else`?

michael-ciniawsky:
```js
if (err) reject(err) // I'm not 100% here, but this should be === (x) => x (return)

resolve(x)
```

```js
if (err) {
   return reject(err)
}

resolve (x)
```

```js
if (err) {
   reject(err)
   // eventually to guard against 'runaway' promises
   return null
}

resolve(x)
```

Please no extra lib, it's not necessary to `promisify`  two callback functions :)

michael-ciniawsky: @ronkorving Thx. Nice catch :)

PR #222, opened 2018-02-08
fix: #217 after update to 4.4.0 TypeError: callbacks[i] is not a func…

RafaPegorari:
…tion


evilebottnawi:
@RafaPegorari Invalid, already have PR https://github.com/webpack-contrib/copy-webpack-plugin/pull/221

PR #223, opened 2018-02-08
chore: fix lint errors in utils/promisify.js

filipesilva:
Fix two lint errors in #221


michael-ciniawsky:
All good I already fixed it on `master` and it's not breaking anything :)

PR #228, opened 2018-02-15
refactor: add pify back using single method call

filipesilva:
Followup to #221


filipesilva:
@evilebottnawi I tested a bit more and this method also doesn't seem to work as expected, I start getting missing private class properties...

Issue #220, opened 2018-02-08
Cannot clone repo on windows


filipesilva:
When trying to clone this repository on windows I get the following error:

```
kamik@T460p MINGW64 /d/sandbox
$ git clone https://github.com/webpack-contrib/copy-webpack-plugin
Cloning into 'copy-webpack-plugin'...
remote: Counting objects: 796, done.
remote: Compressing objects: 100% (34/34), done.
remote: Total 796 (delta 18), reused 31 (delta 13), pack-reused 748
Receiving objects: 100% (796/796), 266.90 KiB | 0 bytes/s, done.
Resolving deltas: 100% (440/440), done.
fatal: cannot create directory at 'tests/helpers/[special?directory]': Invalid argument
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
```

I think the real error is:
```
fatal: cannot create directory at 'tests/helpers/[special?directory]': Invalid argument
```

The special character `?` is not permitted in windows directories.

PR #230, opened 2018-02-22
fix(src/): don't escape non-glob patterns

### LINKED ISSUE #226: Watching single files broken on Windows since 4.4.0 because of wrong path separator
Since version 4.4.0 watching of single files (`from` is no glob and points to file) is broken on Windows. I tracked it down to [this change](https://github.com/webpack-contrib/copy-webpack-plugin/compare/v4.3.1...v4.4.0#diff-a796a48a9ecbc15f8f48dc347832123eR90). It changes the file path on Windows like this:

```js
// pattern before: { absoluteFrom: 'C:\\code\\notes\\src\\sw.js' }
pattern.absoluteFrom = escape(pattern.absoluteFrom);
// pattern after: { absoluteFrom: 'C:/code\\notes\\src\\sw.js' }
```

The path after the change is registered as a file dependency in webpack. But since it's wrong, webpack does not notice any changes to the file.

evilebottnawi:
### `Issues`

- Fixes #226 

No tests, because we should catch this in window CI, but CI don't have this right now :disappointed:

### DIFF
--- a/src/preProcessPattern.js
+++ b/src/preProcessPattern.js
@@ -46,7 +46,8 @@ export default function preProcessPattern(globalRef, pattern) {
         delete fromArgs.glob;
 
         pattern.fromArgs = fromArgs;
-        pattern.absoluteFrom = escape(pattern.context, pattern.from.glob);
+        pattern.glob = escape(pattern.context, pattern.from.glob);
+        pattern.absoluteFrom = path.resolve(pattern.context, pattern.from.glob);
         return Promise.resolve(pattern);
     }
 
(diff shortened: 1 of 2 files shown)


frigus02:
This was harder to test than I thought because of #220 😅 

Still, I can confirm that this PR fixes the referenced issue. I also tested watching a directory and a simple glob and both work fine on Windows. 👍

evilebottnawi:
@michael-ciniawsky we need `absolute` and `glob`, it is impossible fix in `escape`

michael-ciniawsky:
Released in [`v4.4.2`](https://github.com/webpack-contrib/copy-webpack-plugin/releases/tag/v4.4.2)

PR #235, opened 2018-03-01
feat(processPattern): add support for `{RegExp)` matches (`pattern.test`)

jkrems:
We want to match some of our existing asset logic which generates image file names from the source paths. Having a function that generates those paths seemed like the simplest solution.

This mirrors the way the `file-loader` works, allowing to reuse logic. https://github.com/webpack-contrib/file-loader#regexp

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Or, in case of just a `from` with the default destination, you can also use a `{
 |[`fromArgs`](#fromArgs)|`{Object}`|`{ cwd: context }`|See the [`node-glob` options](https://github.com/isaacs/node-glob#options) in addition to the ones below|
 |[`to`](#to)|`{String\|Object}`|`undefined`|Output root if `from` is file or dir, resolved glob path if `from` is glob|
 |[`toType`](#toType)|`{String}`|``|[toType Options](#toType)|
+|[`test`](#test)|`{RegExp}`|``|Pattern for extracting elements to be used in `to` templates|
 |[`force`](#force)|`{Boolean}`|`false`|Overwrites files already in `compilation.assets` (usually added by other plugins/loaders)|
 |[`ignore`](#ignore)|`{Array}`|`[]`|Globs to ignore for this pattern|
 |`flatten`|`{Boolean}`|`false`|Removes all directory references and only copies file names.⚠️ If files have the same name, the result is non-deterministic|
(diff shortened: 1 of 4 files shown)


jkrems:
Actually, trying to make sure that templates are supported in the returned value.

README.md
```diff
@@ -138,6 +138,9 @@ Or, in case of just a `from` with the default destination, you can also use a `{
       from: 'src/'
       to: 'dest/[name].[hash].[ext]',
       toType: 'template'
+      // Optional: provide a regExp to extract additional placeholders
```
michael-ciniawsky: Please make a new 'section'  for this option, or does it only work with `toType: 'template'` ?

michael-ciniawsky: Also needs to be added to `options` table please

jkrems: It will only work with the template `toType`. Will update the table.

jkrems:
Would it make sense to use `pattern.regExp` instead of `pattern.test`? That's consistent with how `file-loader` exposes this feature. Happy to switch to either. :)

jkrems:
* Added to options table
* Renamed to `options.test`
* Added dedicated section for `options.test` to the README containing the same docs that also appear in the `file-loader` README for `options.regExp`

evilebottnawi:
@jkrems rebase :+1:

jkrems:
Rebase'd.

README.md
```diff
@@ -143,6 +144,27 @@ Or, in case of just a `from` with the default destination, you can also use a `{
 ]
 ```
 
+### `test`
+
+Defines a regExp to match some parts of the file path.
```
michael-ciniawsky: `regExp` => `{RegExp}`

README.md
```diff
@@ -143,6 +144,27 @@ Or, in case of just a `from` with the default destination, you can also use a `{
...
+These capture groups can be reused in the name property using `[N]` placeholder.
+Note that `[0]` will be replaced by the entire tested string,
```
michael-ciniawsky:
```

> ℹ️ Note
```

`string` => `{String}`

README.md
```diff
@@ -143,6 +144,27 @@ Or, in case of just a `from` with the default destination, you can also use a `{
...
+whereas `[1]` will contain the first capturing parenthesis of your regex
```
michael-ciniawsky: `regex` => `{RegExp}`

README.md
jkrems: Updated

README.md
jkrems: Went with "entire relative path of the file" while I was touching it since that seemed easier to understand than what the "tested string" was.

README.md
jkrems: Done

jkrems:
Alright, the arguments to the interpolate call changed (relative -> absolute). Updated the test and docs to match that.

michael-ciniawsky:
Released in [`v4.5.0`](https://github.com/webpack-contrib/copy-webpack-plugin/releases/tag/v4.5.0) :tada:

jkrems:
Niiice! That was fast! Thanks a lot for the quick review & release. :)

PR #248, opened 2018-04-02
Fix missing commas in example configs

PandelisZ:
Missing commas in the json for the from to config


evilebottnawi:
@PandelisZ Thanks! Please accept CLA

Issue #229, opened 2018-02-18
[Feature] Only copy file when origin is newer than target on first run


pamtbaau:
Is there a way to only copy files when the `from` is newer than the `to`?

evilebottnawi:
@pamtbaau Thanks for issue, duplicate https://github.com/webpack-contrib/copy-webpack-plugin/issues/169 ?

pamtbaau:
@evilebottnawi Don't think this is a duplicate although they are close...

On each new fresh start of webpack (no watch), files are copied and the origin overwrites the target. I would like changes in the target to be preserved when target is newer then origin. In any case (fresh start of webpack, or a --watch run) I only want the origin to be copied when it is newer than target or target does not exist.

evilebottnawi:
@pamtbaau 
> On each new fresh start of webpack (no watch), files are copied and the origin overwrites the target. I would like changes in the target to be preserved when target is newer then origin.

Not related to copy plugins, it should be done in `webpack` as persistent cache, we just use api webpack https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/writeFile.js#L124, in theory webpack can have cached file and check need output file or not.

And yep duplicate https://github.com/webpack-contrib/copy-webpack-plugin/issues/169, we just don't close issue just for other people, but we can't do nothing here, But if you have good knowledge you can send PR in `webpack` and solve this PR

evilebottnawi:
@pamtbaau only `webpack` have cached fs and contain original mtime, we can do this inside plugin, but it is overload and decrease perf

pamtbaau:
@evilebottnawi I was under the impression that simply comparing mtime of target and origin would be enough to fix the issue. As said, I don't know the ins and outs of webpack... 

Thanks for your time.

evilebottnawi:
@pamtbaau let's left this open for other people, maybe somebody will find simple and good solution

evilebottnawi:
let's close this in favor https://github.com/webpack-contrib/copy-webpack-plugin/issues/169

Issue #169, opened 2017-11-21
[Feature] copy files only on first run


httpeace:
is there any special config settings so the folders are only copied on the first build only? did I miss anything on the documentation

michael-ciniawsky:
By first build you mean when running `webpack` multiple times (without watch mode) ? Files are watched for changes in watch mode, but not between multiple runs atm. Maybe an existence check could be added, but this needs triage first

evilebottnawi:
webpack already do it on own side

evilebottnawi:
It is bad idea we should always to emit files, otherwise build will not be persistent, what could potentially break other plugins (for example clean-webpack-plugin)

hsdonkin:
Running into this issue currently:
- CopyPlugin copies SVG from `./src/svg/*.svg` to another dir like `./components/svg/*.js`
- Tailwind PostCSS plugin is watching `./components/*/**.js` for changes
- Tailwind triggers a build
- CopyPlugin copies SVG.....

alexander-akait:
@hsdonkin exclude svg from watching on Tailwind side, because postcss plugin already watch dev dependecies for Tailwind and will run rebuild if svg was changed

PR #249, opened 2018-04-05
feat: only copy files that have been modified since last webpack invocation

andresa88:
This gives the option to only copy files that have been modified since the last time webpack was. Right now the files are always copied even if they have not been modified. This simply compares file hashes.

Addresses #169 and #229


andresa88:
I'll update the README.md if this is accepted

andresa88:
Cool, thanks for the tip!

PR #256, opened 2018-05-05
feat(preProcessPattern): resolve patterns using node's module resolution algorithm

mmiller42:
I think it would be very useful for this plugin to also allow you to copy assets from Node packages.

It's not as easy as setting `context` to `path.resolve(__dirname, 'node_modules')`. Node has an established [module resolution algorithm](https://nodejs.org/api/modules.html#modules_all_together) that starts at the current directory and traverses up the parents until it either finds the module in a parent's `node_modules` subdirectory or reaches the root directory.

[Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) and [Lerna](https://lernajs.io/) enable developers to manage multiple projects in the same repository and improve dependency installation times by consolidating dependencies that are shared by more than one project in the repo in a `node_modules` directory in the repository root. This is a common example of where a project in a workspace may want to copy an asset from a Node package it has installed, but the asset may exist at `/<pathToRepo>/projects/<project>/node_modules/<package>/<pathToAsset>`, or it might exist at `/<pathToRepo>/node_modules/<package>/<pathToAsset>`.

[`require.resolve`](https://nodejs.org/api/modules.html#modules_require_resolve_request_options), given a request, will resolve with absolute path to the requested asset or throw if the module does not exist in any of the parents' `node_modules` directories.

### DIFF
--- a/src/preProcessPattern.js
+++ b/src/preProcessPattern.js
@@ -61,16 +61,22 @@ export default function preProcessPattern(globalRef, pattern) {
 
     return stat(inputFileSystem, pattern.absoluteFrom)
     .catch(() => {
-        // If from doesn't appear to be a glob, then log a warning
         if (isGlob(pattern.from) || pattern.from.indexOf('*') !== -1) {
+            // If it is a glob, then no worries if it does not exist in the file system
             pattern.fromType = 'glob';
             pattern.glob = escape(pattern.context, pattern.from);
-        } else {
-            const msg = `unable to locate '${pattern.from}' at '${pattern.absoluteFrom}'`;
-            warning(msg);
-            compilation.errors.push(`[copy-webpack-plugin] ${msg}`);
-            pattern.fromType = 'nonexistent';
+        } else if (!path.isAbsolute(pattern.from)) {
+            // Try the node module resolution algorithm to find the module
+            pattern.absoluteFrom = require.resolve(pattern.from);
+            return stat(inputFileSystem, pattern.absoluteFrom);
         }
+    })
+    .catch(() => {
+        // If from doesn't appear to be a glob, then log a warning
+        const msg = `unable to locate '${pattern.from}' at '${pattern.absoluteFrom}'`;
+        warning(msg);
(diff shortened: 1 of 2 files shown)


src/preProcessPattern.js
```diff
@@ -60,18 +60,26 @@ export default function preProcessPattern(globalRef, pattern) {
...
 
     return stat(inputFileSystem, pattern.absoluteFrom)
-    .catch(() => {
-        // If from doesn't appear to be a glob, then log a warning
+    .catch((err) => {
         if (isGlob(pattern.from) || pattern.from.indexOf('*') !== -1) {
+            // If it is a glob, then no worries if it does not exist in the file system
             pattern.fromType = 'glob';
             pattern.glob = escape(pattern.context, pattern.from);
+        } else if (!path.isAbsolute(pattern.from)) {
```
evilebottnawi: What about if i want use absolute path?

mmiller42: On [line 54](https://github.com/webpack-contrib/copy-webpack-plugin/pull/256/files/cde916a4f46cccc0047732834f2a58d73d9af9fc#diff-a796a48a9ecbc15f8f48dc347832123eR54), we specify that if the given `from` path is already absolute, use it as-is.

Then we try to call `stat` on the file. This edit all exists within the `catch` block, which is only evaluated because the inferred absolute path (which is the same as the path you pass in if it was already absolute) did not actually exist.

So if you passed a valid absolute path in for `pattern.from`, the `stat` function call on line 62 wouldn't throw an error.

mmiller42:
@evilebottnawi I'd be happy to add any tests or provide any clarifications if you need, just let me know! I was hoping to take advantage of this (currently I am installing copy-webpack-plugin in my own project using my GitHub repo which, for obvious reasons, isn't ideal, since I have to build it from source after installing).

evilebottnawi:
@mmiller42 Why you close PR? I am one who handle issues/PRs in a lot of repos in `webpack-contrib`. Unfortunately I do not have time all to triangle all. Can you provide minimum repository what shows problems with `actual` and `expected` field here or readme?

PR #258, opened 2018-05-10
Adding the track option

ianwalter:
I strongly believe this option or a similar solution is necessary because if a user is trying to copy a file that is generated during compilation but not tracked as a dependency in their application they will run into an infinite compilation loop. Such is the case with an application I work on which is using the `bundle-tracker-plugin` to generate a sort of bundle manifest file called `webpack-stats.json`. This file is modified a couple of times during the compilation cycle and is not imported into the application so shouldn't be added to `fileDependencies`. This option allows a user to specify not to "track" the files they want copied.


ianwalter:
Apologies I no longer have the time to devote to this.

PR #262, opened 2018-05-22
make sure always have asset information

snadn:
save asset info and add to assets when file hasn't changed

### DIFF
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -84,6 +84,15 @@ export default function writeFile(globalRef, pattern, file) {
                     }
                 );
             }
+            
+            const asset = {
+                size: function() {
+                    return stat.size;
+                },
+                source: function() {
+                    return content;
+                }
+            };
 
             if (!copyUnmodified &&
                 written[file.absoluteFrom] &&
@@ -91,12 +100,14 @@ export default function writeFile(globalRef, pattern, file) {
                 written[file.absoluteFrom]['webpackTo'] === file.webpackTo
             ) {
                 info(`skipping '${file.webpackTo}', because it hasn't changed`);
+                compilation.assets[file.webpackTo] = written[file.absoluteFrom].asset;
                 return;
             } else {
                 debug(`added ${hash} to written tracking for '${file.absoluteFrom}'`);
                 written[file.absoluteFrom] = {
                     hash: hash,
-                    webpackTo: file.webpackTo
+                    webpackTo: file.webpackTo,
+                    asset
                 };
             }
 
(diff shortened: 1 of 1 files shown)


src/writeFile.js
```diff
@@ -84,19 +84,30 @@ export default function writeFile(globalRef, pattern, file) {
...
+                }
+            };
 
             if (!copyUnmodified &&
                 written[file.absoluteFrom] &&
                 written[file.absoluteFrom]['hash'] === hash &&
                 written[file.absoluteFrom]['webpackTo'] === file.webpackTo
             ) {
                 info(`skipping '${file.webpackTo}', because it hasn't changed`);
+                compilation.assets[file.webpackTo] = written[file.absoluteFrom].asset;
```
evilebottnawi: It is break cache, files always emmited

snadn:
@evilebottnawi Can I only use copyUnmodified to resolve the problem of not getting an asset when watching

evilebottnawi:
@snadn why you need this assets?

snadn:
@evilebottnawi when i set plugin like this
```javascript
new CopyWebpackPlugin([
      {
        from: 'node_modules/@cmiot/monitor/lib/monitor_inline.js',
        to: '/',
      },
    ]),
new ScriptExtHtmlWebpackPlugin({
      inline: '/monitor_inline.js'
    }),
```
first i can get assets `monitor_inline` in ScriptExtHtmlWebpackPlugin, but when change, no `monitor_inline` in assets and ScriptExtHtmlWebpackPlugin throw a error;

first info:
image

change info:
image

evilebottnawi:
@snadn we should implement event for this, also it is solve https://github.com/webpack-contrib/copy-webpack-plugin/issues/104 ,feel free to Send PR

evilebottnawi:
Easy example how to create event https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L210

snadn:
@evilebottnawi I'm sorry I'm not very clear about the cache. If there is something wrong with this conclusion, please correct me. In my pr, it will not seem to trigger a copy of the file again by supplying the saved asset to the next build. I see that there is such a piece of code, I do not know if you are referring to the cache processing.

https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L273

Because I am using the windows system, and the project test folder has special characters, so I can not run the test locally

evilebottnawi:
@snadn please create `event` and send a PR, you solution is broke cache, another plugins what use `assets` will always handle this assets and it is bad for perf and may break something places.

evilebottnawi:
@snadn we should use `event` for detect what files emitting, please read docs about how to write plugins and looks on other plugins, i wrote link on example above

snadn:
@evilebottnawi 
In the link you gave https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L176 and the official use case https://webpack.js.org/contribute/writing-a-plugin/#example, the asset information is added directly.

According to my understanding, the compilation.assets should contain the full amount of assets at all times. According to my test, webpack does exactly this: if there are three separate assets a, b, and c, if only a is changed, there will be b and c in the compilation.assets.

The custom hooks should be used to provide additional plug-ins with additional information not available in the compilation. This is in the webpack-manifest-plugin a manifest list not in compilation.

If copy-webpack-plugin wants to control the file name, I think the file should be added as modules instead of assets, using the webpack internal mechanism to handle (this part is not studied in depth, do not know)

PR #284, opened 2018-08-28
feat: add support for target path transform

### LINKED ISSUE #107: [Feature] Manually modify output path

Fetz:
updated version of the PR: webpack-contrib/copy-webpack-plugin/pull/115.

Should resolve #107.

> Why another PR?
>
> Since the other PR is not being updated for some time, and needed to be updated for some time, since I did some differences (mostly to allow Promises with transformPath will need a review)

differences:
- added support to Promises to be aligned with transform (since transform supports Promises)
- updated the tests since there are more test files test files
- updated the Readme to include implementation examples (Function & Promise) like transform in the Readme
- rebased with the latest version

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Or, in case of just a `from` with the default destination, you can also use a `{
 |[`ignore`](#ignore)|`{Array}`|`[]`|Globs to ignore for this pattern|
 |`flatten`|`{Boolean}`|`false`|Removes all directory references and only copies file names.⚠️ If files have the same name, the result is non-deterministic|
 |[`transform`](#transform)|`{Function\|Promise}`|`(content, path) => content`|Function or Promise that modifies file contents before copying|
+|[`transformPath`](#transformPath)|`{Function\|Promise}`|`(targetPath, sourcePath) => path`|Function or Promise that modifies file writing path|
 |[`cache`](#cache)|`{Boolean\|Object}`|`false`|Enable `transform` caching. You can use `{ cache: { key: 'my-cache-key' } }` to invalidate the cache|
 |[`context`](#context)|`{String}`|`options.context \|\| compiler.options.context`|A path that determines how to interpret the `from` path|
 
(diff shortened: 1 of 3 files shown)


EdBoucher:
Any updates on getting this into master? It's extremely useful.

evilebottnawi:
Please rebase PR

evilebottnawi:
@Fetz can you describe use case? Maybe better use `to` as function?

Fetz:
Rebase done

Fetz:
Hi @evilebottnawi 👋,

### My use case:
- I wanted to copy file from a folder to a dist file, but I didn't want part of the path of the file to be included
- I was already using webpack, and I didn't want to add another tool (gulp, grunt or some shell script)

### Why I used `transformPath` and not `to`?
Since there was already a PR open with the functionality and there was ongoing comments with it, and by the comments, I assume that the PR was only blocked, because he need to rebase and update the documentation, and since the original creator, wasn't active working anymore on it I decided to pick his work where he left.

### Why makes more sense (for me) to use `transformPath` instead of `to`
- We have already `transform` logic, so a `transformPath` is consistent with the current options interface
- I assume that most of the times, that the users **target folder** will always be the same but most of the times users will need to transform how the path will be generated inside the **target folder**
- Will be more readable where the files would be copy to (top level of the file tree), and if the user needs more detail of the rest of the path, he can read the transformPath function

as well the original author of the PR included is own take on maybe why should be better:
> Thinking a bit about this, the only benefit of having both templates and a transform is that templates support content hashing. Everything else should be fairly easy to implement in the transform function, if needed.
No idea how often this hash option is used and if it is worth introducing a special transformPath option because of it. 
> @patkec https://github.com/webpack-contrib/copy-webpack-plugin/pull/115#issuecomment-335592368

### Other's people use case

Since I can't talk for everyone, but other people uses case can be found here:
- #293
- #107
- PR https://github.com/webpack-contrib/copy-webpack-plugin/pull/115

Fetz:
Rebase done

Fetz:
@evilebottnawi, maybe @michael-ciniawsky is "ignoring" this PR because is missing the label?
Or there is some higher priority (or complexity related), that other PR's are getting priority?

Is a bit weird that other PR, are being created and being merged faster than this one

evilebottnawi:
@Fetz let's wait 24 hours and merge if not answered, feel free ping me

jomaf1010:
No choice but to fork from @Fetz , still watching for this PR tho

Fetz:
👋 @evilebottnawi pinging you as a reminder regarding the 24 hours time

evilebottnawi:
@Fetz let's merge this release will be after resolve https://github.com/webpack-contrib/copy-webpack-plugin/pull/302 (looks seriously bug)

Fetz:
@evilebottnawi, thanks a lot for merging this

Fetz:
@jomaf1010 you can now use the copy-webpack-plugin@4.6.0, that includes this feature

EdBoucher:
Brilliant, thanks guys!

PR #285, opened 2018-09-06
fix(preProcessPattern): throw an `{Error}` if `pattern.from` is empty or `undefined`

### LINKED ISSUE #278: The plugin should throw an error, if "from" is empty string
Hey guys. 
I spent hours trying to find a reason of weird behavior in my project, but it seems that `copy-webpack-plugin` takes the content of latest-processed file in case if **from** is an empty string, and put it inside the "to:" file. 
I think in such case the plugin should throw an error or show warning at least. 
As I checked in the plugin source, it is related to the wrong path resolving. 

Versions: 
* Webpack: 4.16.2
* copy-webpack-plugin: 4.5.2

Here is the example: 
**webpack.config.js**
```
const webpack = require('webpack');
const CopyWebpackPlugin = require('copy-webpack-plugin');

const config = {
	entry: './index.js',

	plugins: [
		new CopyWebpackPlugin([
			{
				from: '',
				to: './test.json'
			}
		]),
	]
};

module.exports = config;
```

**index.js**
```
console.log('Content from index.js');
```

Webpack run result: 
```
$ ./node_modules/.bin/webpack --mode=development
Hash: 73063a824e1b7c5d19e1
Version: webpack 4.16.2
Time: 1781ms
Built at: 2018-07-26 17:34:11
      Asset      Size  Chunks             Chunk Names
    main.js  3.78 KiB    main  [emitted]  main
./test.json  37 bytes          [emitted]  
Entrypoint main = main.js
[./index.js] 37 bytes {main} [built]
```

After the build content of **test.json** will be: 
```
console.log('Content from index.js');
```

Deliaz:
### Type

- [x] Fix

### Issues

- Fixes #278 

### SemVer

- [x] Patch

### DIFF
--- a/src/preProcessPattern.js
+++ b/src/preProcessPattern.js
@@ -14,6 +14,9 @@ export default function preProcessPattern(globalRef, pattern) {
     pattern = typeof pattern === 'string' ? {
         from: pattern
     } : Object.assign({}, pattern);
+    if (pattern.from === '') {
+        throw new Error('[copy-webpack-plugin] path "from" cannot be empty string');
+    }
     pattern.to = pattern.to || '';
     pattern.context = pattern.context || context;
     if (!path.isAbsolute(pattern.context)) {
--- a/tests/index.js
+++ b/tests/index.js
@@ -227,6 +227,16 @@ describe('apply function', () => {
 
             expect(createPluginWithNull).to.throw(Error);
         });
+
+        it('throws an error if the "from" path is an empty string', () => {
+            const createPluginWithNull = () => {
+                CopyWebpackPlugin({
+                    from: ''
+                });
+            };
+
+            expect(createPluginWithNull).to.throw(Error);
+        });
     });
 
     describe('with glob in from', () => {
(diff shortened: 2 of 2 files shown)


src/preProcessPattern.js
```diff
@@ -14,6 +14,9 @@ export default function preProcessPattern(globalRef, pattern) {
     pattern = typeof pattern === 'string' ? {
         from: pattern
     } : Object.assign({}, pattern);
+    if (pattern.from === '') {
```
michael-ciniawsky:
```js
if (!pattern.from) {
 throw new Error(
  'Copy Plugin\n\npattern.from must be defined\n\n(https://github.com/webpack-contrib/copy-webpack-plugin#from)\n'
 )
}

```

Deliaz:
Just wondering, do I need to finish something? Not sure about the workflow here.

Deliaz:
@michael-ciniawsky Hi, am I right that you need to sign CLA? :)

evilebottnawi:
@Deliaz thanks

Deliaz:
> @Deliaz thanks

Thanks. But I am still not sure, is there anything that is required from my side?

evilebottnawi:
@Deliaz no, we can merge this

PR #290, opened 2018-09-28
fix(processPattern): add `glob` directory context to `contextDependencies`

alan-agius4:
At the moment when using globs with `watch` mode and files are added after the initial compilation, they are not copied because the globbed `dirname` is not added to the context


alan-agius4:
Can we get this release please?

michael-ciniawsky:
Released in [`v4.5.3`](https://github.com/webpack-contrib/copy-webpack-plugin/releases/tag/v4.5.3) :tada:

Issue #294, opened 2018-10-05
Excluding single files


ivansieder:
The Vue CLI offers the possibility to use the public/ directory, which will copy it's files 1:1 to the dist/ folder when it's being built. That's what copy-webpack-plugin is being used for.

Now additionally, some of the .html-files from public/ are being are being used just as templates for html-webpack-plugin to auto-inject scripts. That means, that these files don't need to be copied to the final dist/ folder. The issue which arises now is, that using the "ignore" property for this won't work all too well, as using "index.html" for example would mean that it's using "index.html" as a glob and therefor excluding all index.html files.

Is there any way to exclude single files with relative paths (i.e. "public/index.html", "public/other/index.html" etc. or what would be the correct approach to handle these cases?)

evilebottnawi:
@ivansieder interesting, can you provide minimum reproducible repo with example what you actually have and expected. Why `ignore` with `index.html` is not suitable for you?

ivansieder:
@evilebottnawi repo could be used this one: https://github.com/ivansieder/vue-cli-nested-index-files-bug. Right now, vue cli is passing ["index.html", ".DS_Store"] to the copy-webpack-plugin, which means, that the glob is being matched with index.html, which in turn means, that public/index.html, public/other/index.html, public/some/extremely/nested/one/index.html all would be ignored. That's why ignore in this case won't work really rell in my opinion (maybe there's a way around that which I don't know yet).

The repo above should be installed and inside node_modules, there's the @vue/cli-service/lib/config/app.js file and at the really bottom, there's the copy-webpack-plugin being configured. Does that help or do you need some additional information?

evilebottnawi:
@ivansieder what about implement `ignore` as function?

ivansieder:
@evilebottnawi what could such a function possibly look like?

evilebottnawi:
@ivansieder depends what information you need to detect should you ignore or not

ivansieder:
@evilebottnawi just talking about vue cli (but I think this could be useful in other cases, too), a function which gets called with the current `file` should be more than enough. that way, inside the function, you could check if that path matches one you have in an array or somewhere else and return true or false? I would however decouple it from the "normal" `ignore`, as it could happen (that's also the case for vue cli) that you need something being ignored by globs and single files.

```javascript
webpackConfig
        .plugin('copy')
          .use(require('copy-webpack-plugin'), [[{
            from: publicDir,
            to: outputDir,
[...]

PR #295, opened 2018-10-05
feat(ignoreFiles): added an ignore function to allow granular file exclusions

ivansieder:
Refers to #294


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "copy-webpack-plugin",
-  "version": "4.5.2",
+  "version": "4.6.0",
```
evilebottnawi: Revert this, changelog and lock file - this do contributors

src/index.js
```diff
@@ -92,17 +93,17 @@ function CopyWebpackPlugin(patterns = [], options = {}) {
...
 
             Promise.all(tasks)
-            .catch((err) => {
-                compilation.errors.push(err);
-            })
-            .then(() => callback());
+                .catch((err) => {
+                    compilation.errors.push(err);
+                })
+                .then(() => callback());
```
evilebottnawi: Don't modify code style in this PR, this will be do in next update code based

src/preProcessPattern.js
```diff
@@ -60,43 +65,43 @@ export default function preProcessPattern(globalRef, pattern) {
...
+                pattern.glob = escape(pattern.absoluteFrom);
+                pattern.fromArgs = {
+                    dot: true
+                };
+                fileDependencies.push(pattern.absoluteFrom);
+            } else if(!pattern.fromType) {
+                info(`Unrecognized file type for ${pattern.from}`);
+            }
+            return pattern;
+        });
```
evilebottnawi: Avoid modify code style

src/writeFile.js
```diff
@@ -11,109 +11,109 @@ export default function writeFile(globalRef, pattern, file) {
...
+                    info(`writing '${file.webpackTo}' to compilation assets from '${file.absoluteFrom}'`);
+                    compilation.assets[file.webpackTo] = {
+                        size: function() {
+                            return stat.size;
+                        },
+                        source: function() {
+                            return content;
+                        }
+                    };
+                });
```
evilebottnawi: Avoid modify code based

evilebottnawi:
@ivansieder why don't use `ignore` option?

ivansieder:
Well but that would mean that if I want to exclude single files and a glob pattern, that I have to write custom logic for handling glob patterns, which is pretty difficult IMHO

evilebottnawi:
@ivansieder can you provide example?

ivansieder:
@evilebottnawi an example could be that someone wants to (simplified) ignore all .DS_STORE files which appear somewhere but also wants to exclude explicitly some single index.html files from being copied (that's taken from the Vue CLI example).

evilebottnawi:
@ivansieder what about allow ignore be `String/Function/Array<String|Function>` (i.e. allow be function and allow be function in array)?

evilebottnawi:
@ivansieder sounds reasonable, maybe rename `ignoreFn` to `ignoreFiles` to avoid misleading and in next major release rename `ignore` to `ignoreGlob`

ivansieder:
Done :)

src/processPattern.js
```diff
@@ -73,6 +73,19 @@ export default function processPattern(globalRef, pattern) {
                 }
             }
 
+            // Run ignore functions
+            for (let i = 0; i < pattern._ignoreFunctions.length; i++) {
+                debug(`running ignore function against ${file.relativeFrom}`);
+                if (
+                    pattern._ignoreFunctions[i]({
+                        absolutePath: file.absoluteFrom,
+                        relativePath: file.relativeFrom
```
evilebottnawi: I think better arguments is `absolutePath` and `context`, it is allow to know `relative` path and also allow get context

evilebottnawi:
@ivansieder no commits :smile:

src/processPattern.js
ivansieder: I've named it absolute and relative path, as that's mostly related to absoluteFrom and relativeFrom.

src/preProcessPattern.js
```diff
@@ -48,6 +48,11 @@ export default function preProcessPattern(globalRef, pattern) {
   pattern.context = normalizePath(pattern.context);
   pattern.ignore = globalRef.ignore.concat(pattern.ignore || []);
 
+  // One global and one pattern-specific ignore function is allowed
+  pattern._ignoreFunctions = [];
+  if (globalRef.ignoreFiles && typeof globalRef.ignoreFiles === 'function') pattern._ignoreFunctions.push(globalRef.ignoreFiles);
```
evilebottnawi: Avoid `globalRef`, `ignore` and `context` on global level will be removed in next major release in favor `pattern.ignore` and `pattern.context`

src/processPattern.js
```diff
@@ -90,6 +90,21 @@ export default function processPattern(globalRef, pattern) {
             logger.debug(`${glob} doesn't match ${file.relativeFrom}`);
           }
 
+          // Run ignore functions
+          for (let i = 0; i < pattern.allIgnoreFunctions.length; i++) {
```
evilebottnawi: Use `forEach`

src/preProcessPattern.js
ivansieder: so there should be no global pattern possible anymore? just on the pattern directly?

evilebottnawi: yes

evilebottnawi:
Also please provide use case where you need `ignore` as glob pattern and `ignoreFiles` as function, because logic is misleading and no need add new option as i say above

ivansieder:
@evilebottnawi would you mind giving it another look?

> 
> 
> Also please provide use case where you need `ignore` as glob pattern and `ignoreFiles` as function, because logic is misleading and no need add new option as i say above

One use case is the Vue.js CLI 3. For multi-page applications, there might be multiple subdirectories, where an index.html needs to be ignored. Using the glob pattern doesn't work in this case, as not all index.html files have to be ignored.

src/processPattern.js
ivansieder: i've chosen the for loop, as i need to return the resolved promise, if one pattern matches. can I leave it like that for this use case?

evilebottnawi:
@ivansieder Please provide example, I still can not understand why we need a new option when we can implement `ignore` as `Function`

ivansieder:
> 
> 
> @ivansieder Please provide example, I still can not understand why we need a new option when we can implement `ignore` as `Function`

It WOULD actually be possible to use `ignore` as a function, but does that actually make sense? That would mean, that a user, who wants to use the glob pattern needs to evaluate the glob himself. Or how could a design, which combines the glob pattern and a function, look like? What's your idea?

evilebottnawi:
@ivansieder you should install `minimatch` and check this in function, we can't cover all edge cases so we provide function to allow any difficult ignore cases

ivansieder:
@evilebottnawi I'm afraid I can't follow you now anymore. What exactly should/could this function look like, or does that even make sense? Otherwise I'll proceed with closing this PR and report that back to the Vue.js-related issue for now.

evilebottnawi:
we should implement `ignore` as `Function`

ivansieder:
you mean unifying ignore and ignoreFiles? so that way someone would have to do the minimatch on his own and this would introduce a breaking change. is that correct?

evilebottnawi:
@ivansieder yes, if you use `ignore` as `Function` you need own logic for ignore (you can use any libraries `minimatch`, `anymatch`, `RegExp` and etc)

ivansieder:
alright so the proposal would be to let `ignore` be either an array of globs or a function and if it's an array, preserve the old behavior and otherwise use the function?

evilebottnawi:
@ivansieder yes :+1:

ivansieder:
alright, and if the global setting is an array of globs, probably also the pattern-specific ignore value should be an array of globs? or should it be possible to have an array of globs globally and a function as pattern-specific value and vice-versa?

evilebottnawi:
I think we can avoid global option because it will be removed in next major release

ivansieder:
The purpose for the Vue.js case has been solved differently.

PR #296, opened 2018-10-10
fix(processPattern): don't add `'glob'` as directory when it is a file (`contextDependencies`)

alan-agius4:
In some cases we might have someone like this which in that cause we wouldn't want to add the parent directory as a webpack context dependencies
```
{
  ignore: [ '.gitkeep', '**/.DS_Store', '**/Thumbs.db' ],
  from: { glob: 'favicon.ico', dot: true },
  toType: 'dir',
  fromType: 'glob'
}
```

### DIFF
--- a/src/processPattern.js
+++ b/src/processPattern.js
@@ -1,5 +1,6 @@
 import globby from 'globby';
 import pLimit from 'p-limit';
+import isGlob from 'is-glob';
 import path from 'path';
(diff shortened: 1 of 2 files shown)


michael-ciniawsky:
Released in [`v4.5.4`](https://github.com/webpack-contrib/copy-webpack-plugin/releases/tag/v4.5.4) 🎉

PR #302, opened 2018-10-25
fix: handle undefined and null as stats value

alan-agius4:
In certain cases depending on the inputFileSystem `stats` doesn't throw, but returns a falsey value such as `undefined` and `null`. In this case an error from globby will be thrown `Patterns must be a string or an array of strings`

Relates to: https://github.com/angular/angular-cli/issues/12686


alan-agius4:
@evilebottnawi, done

alan-agius4:
Oh, you need to support Node 4. Il'll update the code.

PR #308, opened 2018-12-01
docs(README): fix function name `transformPath` in code sample

esphas:
The example for `transformPath`(`Promise`) uses `transform`, fixed that.


evilebottnawi:
Thanks!

PR #309, opened 2018-12-03
Support for manifest plugin and other modules utilizing moduleAsset hook

skliffmueller:
It's stupid simple. Announce to moduleAsset hook. Now original reference name links to new asset name. I'm sure this will help resolve other related conflicts with plugins and post processing against asset paths.

Issue #104

### DIFF
--- a/src/writeFile.js
+++ b/src/writeFile.js
@@ -117,6 +117,12 @@ export default function writeFile(globalRef, pattern, file) {
                 return;
             }
 
+            // Used to announce these files as moduleAssets.
+            // This allows the paths to be resolved by things like the manifest plugin
+            compilation.hooks.moduleAsset.call({
+                userRequest: file.relativeFrom
+            }, file.webpackTo);
+
             info(`writing '${file.webpackTo}' to compilation assets from '${file.absoluteFrom}'`);
             compilation.assets[file.webpackTo] = {
                 size: function() {


src/writeFile.js
```diff
@@ -117,6 +117,12 @@ export default function writeFile(globalRef, pattern, file) {
                 return;
             }
 
+            // Used to announce these files as moduleAssets.
+            // This allows the paths to be resolved by things like the manifest plugin
+            compilation.hooks.moduleAsset.call({
+                userRequest: file.relativeFrom
+            }, file.webpackTo);
```
evilebottnawi: Also not sure it is right solution, because we interpreted files as module assets

evilebottnawi: @skliffmueller Here (https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L59) we need implement hook `compilation.applyPluginsAsync('webpack-manifest-plugin-before-emit', callback);`

And use this hook in this plugin.

We should not pass this as `module` because it can break analyze tools.

evilebottnawi: And to avoid problems with other plugins we need implement new hooks in `webpack`, example `manifest` which call after `emit`. It is also easy implement and feel free to ping me in `webpack` repo

skliffmueller: I don't think that will help the situation, as copy-webpack-plugin doesn't need to know what the manifest looks like, webpack-manifest-plugin needs to know what copy-webpack-plugin has modified. In webpack-manifest-plugin stats.assets from the complication object is the only point where the copy-webpack-plugin assets exist. and moduleAssets is the only object map to reference the hash file names to the original path. Otherwise it defaults to the file name.
https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L88-L114

So either copy-webpack-plugin is not implemented correctly, webpack complication assets is not implemented correctly (needs to include ability to set asset name and path), or webpack-manifest-plugin isn't implemented correctly and needs to listen to an additional hook to build against the object mapping between hash file names and original file names.

evilebottnawi: > needs to listen to an additional hook to build against the object mapping between hash file names and original file names.

Yes, because adding this as `module` can break analyze tools, also it is not module.

Manifest plugin have 2 problem:
1. Using `emit` hooks (it is not safe because other plugins also add assets on this hook and some assets may disappear). As i said above we need new hook in `webpack` for this. You can send a PR in `webpack`, I will support you, should be easy.
2. Plugin should add hook for adding assets (I showed the place where it should be).

These tasks are in my todo lists, but at the moment there are more priorities, so you can start doing this and I will be happy to help.

skliffmueller: Okay I might not have clearly understood you at first. But I get it, we are on the same page. I will look into doing this over the weekend when work isn't chasing after me.

evilebottnawi:
Fixed in master, now we have `sourceFilename` in assets info, so you can get easy get original filename, maybe you need to pen an issue for supporting it in the plugin for generation these

PR #316, opened 2019-01-23
Copy file permissions

jeepingben:
Reviving PR #119
Would resolve issue #35
This is updated to build against the latest master.  Just as in the original PR I have not been able to add unit tests because I can't stat the emitted files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Or, in case of just a `from` with the default destination, you can also use a `{
 
 **webpack.config.js**
 ```js
-[`
+[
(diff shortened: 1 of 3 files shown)


anandbaskaran:
When do you think this will be merged? Is there any other options in the meanwhile?

evilebottnawi:
@anandbaskaran please read comment above

jeepingben:
This has drifted quite a bit since January and I don't think I will be able to make it correct + get the needed PRs into webpack.

PR #337, opened 2019-02-18
fix: 🐛 normalize path before check is glob

### LINKED ISSUE #317: isGlob return false for windows path with backslashes
As the case we have `pattern.from = 'C:\my-repo\{manifest,manifest}.json'` in windows and `pattern.from = '/my-repo/{manifest,manifest}.json'` in linux. 

https://github.com/webpack-contrib/copy-webpack-plugin/blob/adf10462cf04b75a8ad59a42ceacf3b033850a8a/src/preProcessPattern.js#L67

It returned `true` in linux but returned `false` in windows because of https://github.com/micromatch/is-glob/issues/10.

Tried:
```node
> require('is-glob')('C:\\{foo,bar}.js')
false
> require('is-glob')('/{foo,bar}.js')
true
```

I also tried `is-glob@2.0.1` and got `true` in both cases.

Any idea to make it works in Windows? Thanks!

https://github.com/egoist/poi/issues/533

VdustR:
normalize the path for wildcard in windows

Issues: fix #317

This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fix #317 

### Breaking Changes

_n/a_

### Additional Info

fixed in the same way with https://github.com/egoist/poi/issues/533


package.json
```diff
@@ -40,14 +40,15 @@
...
     "is-glob": "^4.0.0",
     "loader-utils": "^1.1.0",
     "minimatch": "^3.0.4",
-    "p-limit": "^2.1.0"
+    "normalize-path": "^3.0.0",
+    "p-limit": "^2.1.0",
+    "serialize-javascript": "^1.4.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.1.5",
```
evilebottnawi: Don't touch this file in this PR

VdustR: is it okay that eslint warning `import/no-extraneous-dependencies`?

evilebottnawi: @VdustR on what?

VdustR: https://github.com/webpack-contrib/copy-webpack-plugin/blob/6457c764334e3de36e86a5df8b88b9a52f24f6d3/src/preProcessPattern.js#L4

evilebottnawi: Oh, all good, my mistake, tired

evilebottnawi:
Please accept CLA

VdustR:
> Please accept CLA

signed

src/preProcessPattern.js
```diff
@@ -30,6 +31,18 @@ export default function preProcessPattern(globalRef, pattern) {
...
+    pattern.from = normalizePath(pattern.from);
+  } else if (
+    typeof pattern.from === 'object' &&
+    typeof pattern.from.glob === 'string'
+  ) {
+    pattern.from = {
+      ...pattern.from,
+      glob: normalizePath(pattern.from.glob),
+    };
+  }
```
evilebottnawi: Find one interesting thing https://github.com/floatdrop/gulp-watch/blob/master/index.js#L54 , we need add test for '!'

evilebottnawi: Maybe best place for this here https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/utils/escape.js?

VdustR: Sorry, I didn't get it.

Do you mean test `'[!]\\hello.txt'`?

evilebottnawi: Something like this `!(directory)/**/*.txt`

Issue #349, opened 2019-03-06
Treat [hash] and [contenthash] correctly


rudionrails:
* Operating System: OSX
* Node Version: 11.3
* NPM Version: 6.7.0
* webpack Version: 4+
* copy-webpack-plugin Version: 5.0.0

### Feature Proposal

I would like to treat `[hash]` and `[contenthash]` correctly in the file/directory string. This is related to [this](https://github.com/webpack-contrib/copy-webpack-plugin/issues/237) issue, but [other](https://github.com/webpack-contrib/mini-css-extract-plugin) [packages](https://github.com/jantimon/html-webpack-plugin) have already implemented this accordingly. PR will follow shortly - looking forward to feedback.

### Feature Use Case

Keeps handling consecutive builds separated. Mostly used to move files into a common sub-directory:

```
d41d8c/main.js
d41d8c/index.html
ax798f/main.js
ax798f/index.html
```

evilebottnawi:
@rudionrails What do you mean Treat [hash] and [contenthash] correctly? `contenthash` is `hash` of content and it is already works as expected, `hash` is deprecated and will be removed in next version in `webpack` and in plugin, it is for backward compatibility

evilebottnawi:
Also solution here https://github.com/rudionrails/copy-webpack-plugin/commit/ef2b2df6bf09b9f20fdfc1f7585cc4c4aceca21e is invalid and broken long term cache in some cases

evilebottnawi:
Just information for generate hash/contenthash respect `loaderUtils.interpolateName`

rudionrails:
@evilebottnawi thanks for the swift response. Where can I find that [hash] will be deprecated in webpack? I did not find this.

rudionrails:
To my knowledge, this has just been added in v4-alpha https://webpack.js.org/configuration/output/#outputhashfunction

rudionrails:
Also, loader-utils cannot handle it, as it has no access to webpack `compilation`, hence it is added inside this repo. I made a comment in the code about this.

evilebottnawi:
@rudionrails we should fix it in `loader-utils` not here, no public information about `[hash]` right now, but many docs already use `[contenthash]`. Relevant issue https://github.com/webpack/webpack/issues/8535, we should implement support `hash*` options in `loader-utils`, no need hack here, feel free to send a PR to `loader-utils`, it should be easy

evilebottnawi:
[...]

PR #350, opened 2019-03-06
feat: treat [hash] and [contenthash] as expected in webpack (#349)

rudionrails:
BREAKING CHANGE: [hash] is treated as webpack (v4+) build hash. [contenthash] remains as expected.

This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Webpack distinguishes between hash and contenthash as explained here: https://github.com/webpack-contrib/copy-webpack-plugin/issues/349

### Breaking Changes

This change breaks backwards compatibility with people using [hash], but actually mean [contenthash]. The library used to generate this ([loader-utils](https://www.npmjs.com/package/loader-utils)) 
only keeps [hash] for backwards compatibility themselves as described [here](https://github.com/webpack/loader-utils/blob/master/lib/interpolateName.js#L91).

### DIFF
--- a/src/postProcessPattern.js
+++ b/src/postProcessPattern.js
@@ -2,6 +2,7 @@ import path from 'path';
 
 import crypto from 'crypto';
 
+import webpack from 'webpack';
 import loaderUtils from 'loader-utils';
 import cacache from 'cacache';
 import serialize from 'serialize-javascript';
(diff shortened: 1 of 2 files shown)


rudionrails:
Closing as per discussion here #349

PR #355, opened 2019-03-12
[DNM] fix: watching on windows

### LINKED ISSUE #347: Watch mode not working correctly for certain filetypes
* Operating System: Windows 10
* Node Version: 10.12.0
* NPM Version: 6.4.1
* webpack Version: 4.29.6
* copy-webpack-plugin Version: 5.0.0

### Expected Behavior
- Files that are copied by this plugin should be watched & recopied when they are modified.
- Copied directories should watch for new/deleted files
- Globs should watch for new files matching it. The [`copy-and-watch`](https://github.com/HoldYourWaffle/copy-and-watch/tree/4632231c33e9d81987c5d45fb1c93638206d9352) npm module does this already by watching the parent directory and copying any files that match the glob. (Linking to a specific commit in my own fork because it actually works there, and in my opinion it's more readable before I had to reformat because of lint settings.)

### Actual Behavior
At first I thought copied files weren't watched at all, but after some headaches I found out that it's just files without a webpack loader. `json` files are copied, [`ejs`](https://ejs.co/) files aren't.

To add salt to injury, in my own project these "no-loader" files got copied *sometimes* when I modified another file that did have a loader first and if they were copied by a glob, and then it still only worked once for some reason. However I wasn't able to reproduce this behavior in my reproducing-repository (it might have something to do with the `context` option). For now I think we should focus on making it work in a not-complicated environment, if the issue persists after that I'll look into it more extensively.

As far as I can tell the `copyUnmodified` option doesn't change anything.

### Code
I've created a reproducing repository [here](https://github.com/HoldYourWaffle/webpack-copy-issue), for my own project is private.

### How Do We Reproduce?
- Clone aforementioned repository
- Run `npx webpack --watch`
- Modify one of the `ejs` files or create a new file in the `assets` directory
- Observe the absence of a build trigger

evilebottnawi:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #347

### DIFF
--- a/src/postProcessPattern.js
+++ b/src/postProcessPattern.js
@@ -6,7 +6,6 @@ import loaderUtils from 'loader-utils';
 import cacache from 'cacache';
 import serialize from 'serialize-javascript';
 import findCacheDir from 'find-cache-dir';
-import normalizePath from 'normalize-path';
 
 import { name, version } from '../package.json';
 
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Fixed

PR #360, opened 2019-03-21
fix: persist assets between rebuilds

### LINKED ISSUE #356: After recompiling plugin does not add files into assets-manifest.json
* Operating System: Any
* Node Version: 11.10.0
* NPM Version: 6.7.0
* webpack Version: 4.29.6
* copy-webpack-plugin Version: 5.0.0

### Expected Behavior

Expect to have all copied assets to be in assets-manifest.json

### Actual Behavior
After recompiling there are no any assets which is copied by plugin
Here is plugin settings:
```js
new CopyWebpackPlugin([
            { from: './.env.development' },
            { from: './src/common/assets/manifest.json' },
            { from: './src/common/assets/*.png', flatten: true },
            { from: './src/common/assets/favicon.ico' },
            { from: './src/common/assets/webPush/pushPackage.zip' },
            { from: './node_modules/react/umd/react.development.js' },
            { from: './node_modules/react-dom/umd/react-dom.development.js' },
]),
```

After first compilation it has such content
```json
{
  "main.js": "/main.js",
  ".env.development": "/.env.development",
  "android-launchericon-144-144.png": "/android-launchericon-144-144.png",
  "android-launchericon-192-192.png": "/android-launchericon-192-192.png",
  "android-launchericon-48-48.png": "/android-launchericon-48-48.png",
  "android-launchericon-512-512.png": "/android-launchericon-512-512.png",
  "android-launchericon-72-72.png": "/android-launchericon-72-72.png",
  "android-launchericon-96-96.png": "/android-launchericon-96-96.png",
  "favicon-16x16.png": "/favicon-16x16.png",
  "favicon-32x32.png": "/favicon-32x32.png",
  "favicon.ico": "/favicon.ico",
  "ios-appicon-1024-1024.png": "/ios-appicon-1024-1024.png",
  "ios-appicon-120-120.png": "/ios-appicon-120-120.png",
  "ios-appicon-152-152.png": "/ios-appicon-152-152.png",
  "ios-appicon-180-180.png": "/ios-appicon-180-180.png",
  "ios-appicon-76-76.png": "/ios-appicon-76-76.png",
  "ios-launchimage-1024-768.png": "/ios-launchimage-1024-768.png",
  "ios-launchimage-1242-2208.png": "/ios-launchimage-1242-2208.png",
  "ios-launchimage-1334-750.png": "/ios-launchimage-1334-750.png",
  "ios-launchimage-1536-2048.png": "/ios-launchimage-1536-2048.png",
  "ios-launchimage-2048-1536.png": "/ios-launchimage-2048-1536.png",
  "ios-launchimage-2208-1242.png": "/ios-launchimage-2208-1242.png",
  "ios-launchimage-640-1136.png": "/ios-launchimage-640-1136.png",
  "ios-launchimage-640-960.png": "/ios-launchimage-640-960.png",
  "ios-launchimage-750-1334.png": "/ios-launchimage-750-1334.png",
  "ios-launchimage-768-1024.png": "/ios-launchimage-768-1024.png",
  "manifest.json": "/manifest.json",
  "plucky.mp3": "/plucky.a90b8d5c.mp3",
  "pushPackage.zip": "/pushPackage.zip",
  "react-dom.development.js": "/react-dom.development.js",
  "react.development.js": "/react.development.js"
}
```
and after recompilation:
```json
{
  "main.js": "/main.js",
  "android-launchericon-192-192.png": "/android-launchericon-192-192.e17da0a1.png",
  "plucky.mp3": "/plucky.a90b8d5c.mp3"
}
```
but copied assets are present in dist folder

изображение

Unfortunately I dont know why webpack recompiles it 3 times without code changing :(

### Code

[Gist link](https://gist.github.com/budarin/c41a942684f90d7fc10ac2531dfcbeb2)

chrisblossom:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Copied files are still part of the webpack assets between rebuilds.

After revamping [clean-webpack-plugin](https://github.com/johnagan/clean-webpack-plugin) to automatically remove unused webpack assets we've started getting reports of files being removed that were copied via this plugin. Since files copied are still part of webpack's assets, they should be persistent between rebuilds.

I am unsure how to write tests for this with your current test setup. If this is wanted, please point me in the right direction how to properly do this.

I think this also fixes #356 and possibly closes #309.

### Breaking Changes

I do not think this is a breaking change.

### DIFF
--- a/src/postProcessPattern.js
+++ b/src/postProcessPattern.js
@@ -155,6 +155,16 @@ export default function postProcessPattern(globalRef, pattern, file) {
       .then((content) => {
         const hash = loaderUtils.getHashDigest(content);
 
+        // persist assets between rebuilds
+        compilation.assets[file.webpackTo] = {
+          size() {
+            return stats.size;
+          },
+          source() {
+            return content;
+          },
+        };
+
         if (
           !copyUnmodified &&
           written[file.webpackTo] &&
@@ -189,15 +199,6 @@ export default function postProcessPattern(globalRef, pattern, file) {
             file.absoluteFrom
           }'`
         );
-
-        compilation.assets[file.webpackTo] = {
-          size() {
-            return stats.size;
-          },
-          source() {
-            return content;
-          },
-        };
       });
   });
 }


chrisblossom:
Sorry, guess I'm too used to tests being ran in git pre-push and thought everything was passing. Looking more into getting these tests to pass, although it looks difficult because of how they are currently designed.

In the meantime, please advise if this PR has a chance to be accepted so I don't waste too much time on fixing the tests if it is for nothing.

Thanks!

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [Unreleased]
+
+
+### Bug Fixes
+
+* persist assets between rebuilds
+
+
```
evilebottnawi: Avoid this change

evilebottnawi:
We don't add new files if their content is same before previous build (example in watch), we need more information to fix this problem

chrisblossom:
> We need tests for this to avoid future regressions, also it is break copyUnmodified, can you create minimum reproducible test repo?

I agree. Because your current test rely on the `assets`, I'll need to figure out how to properly update the tests and will also look into `copyUnmodified`.

> We don't add new files if their content is same before previous build (example in watch), we need more information to fix this problem

The issue is the files previously added are not being recorded in webpack's asset list on rebuild. The asset list is not a list of files that have been updated, but rather all files in the current build. For example, lets say you have the following webpack files:

```
dist/
    1.bundle.js
    bundle.js
    static-file.jpg // copied with copy-webpack-plugin
```

during the first webpack build, webpack's assets will equal (a simplified version):
```js
['1.bundle.js', 'bundle.js', 'static-file.jpg']
```

Lets say whatever source file made up `1.bundle.js` was modified, webpack's assets will now equal:
```js
['1.bundle.js', 'bundle.js'] // static-file.jpg is missing
```

Notice how even though the source files that made up `bundle.js` have not been modified, it is still on webpack's asset list. Because webpack's assets are supposed to be (at least how I understand it) a list of all current webpack assets (regardless if they've been updated or not) I think files copied via `copy-webpack-plugin` should always be included in webpack's assets on rebuilds.

This does not mean the file needs to be copied again, but instead reported to webpack's asset list (which this PR does).

CHANGELOG.md
chrisblossom: change removed

chrisblossom:
Looking more into the source code, it looks like `compilation.assets[file.webpackTo]` is where webpack is told to copy the file to and from? If that is the case, I am unsure how to solve this issue or if it is even possible without telling webpack to always copy the file.

https://github.com/webpack-contrib/copy-webpack-plugin/blob/95cf57e3418baf47efcdc216f82f4b9c0d02cc94/src/postProcessPattern.js#L193-L200

evilebottnawi:
@chrisblossom yep, we need new hook for this, it is also allow support manifest plugins, and we should implement this in webpack. I think we need implement even 2 hooks:
- before-assets-manifest (we add array copied files in this stage) (sync)
- assets-manifest (manifest plugins generate manifests) (async)

So it is allow to us don't emit files between build (if they don't changed) and allow to clean plugin get copied assets. As workaround we can introduce this hooks on plugin level (before we implement this in webpack)

chrisblossom:
@evilebottnawi Are you sure webpack doesn't handle the file caching / doesn't copy unchanged files internally? Maybe @sokra or @TheLarkInn can better answer this.

Also, is the purpose of not copying an unchanged file for performance? Is copying a file slower than reading it and calculating the checksum?

evilebottnawi:
@chrisblossom 

> @evilebottnawi Are you sure webpack doesn't handle the file caching / doesn't copy unchanged files internally? Maybe @sokra or @TheLarkInn can better answer this.

hm, need investigate, this code was from other contributors (legacy) https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L316

Issue #261, opened 2018-05-19
Conflict with clean-webpack-plugin?


kevinzheng:
When I use `clean-webpack-plugin` with `copy-webpack-plugin`, I got `ERROR in [copy-webpack-plugin] unable to locate 'dist/manage.bundle.js' at '/Users/XXX/Workspace/project/dist/manage.bundle.js'`, but when I remove `clean-webpack-plugin`, it works fine, what the problem is?

```javascript
    plugins: [
        new CleanWebpackPlugin(['dist']),
        new UglifyJSPlugin({
            sourceMap: true
        }),
        new CopyWebpackPlugin([
            {
                from: 'dist/manage.bundle.js',
                to: '../public/js/'
            }
        ], { debug: 'info' })
    ],
```

evilebottnawi:
@kevinzheng because `CleanWebpackPlugin` remove `dist/manage.bundle.js`, it is expected

evilebottnawi:
@kevinzheng i don't understand your, you remove files and when want to copy this file?

kevinzheng:
@evilebottnawi , I want to `clean-webpack-plugin` before the compiling, and `copy-webpack-plugin` after the compiling, ehh... I want to clean the `dist` dir before building, and  copy the building result to another dir.

evilebottnawi:
@kevinzheng out of scope `copy-webpack-plugin`, use any `copy` plugin to move something after build process.

cfkxzsat:
same here..

mycoin:
``` javascript
new CleanWebpackPlugin({
        cleanStaleWebpackAssets: false, // resolve conflict with `CopyWebpackPlugin`
}),
```

evilebottnawi:
Somebody can create reproducible test repo?

chrisblossom:
The issue is `copy-webpack-plugin` does not persist the assets copied in webpack's current asset list. I attempted to resolve this issue in #360 but the PR looks to be stale. If a maintainer can provide some insight on how to move that PR forward I would be willing to make the necessary changes to solve this.

evilebottnawi:
@chrisblossom this PR is breaking change, we will fix it in next release

tobia:
`cleanStaleWebpackAssets: false` does not solve the problem for me.

I use CopyWebpackPlugin to copy static files into the bulid folder:

```
  plugins: [
    new CleanWebpackPlugin(),
    new MiniCssExtractPlugin({
[...]

PR #372, opened 2019-04-17
Improve clarity on how plugin works with built files

soundasleep:
It appears that copy-webpack-plugin is designed to copy files _outside_ of the build process, rather than copy files _created by_ the build process. See: https://github.com/webpack-contrib/copy-webpack-plugin/issues/261

If one uses webpack-copy-plugin to try and copy a built file, if the file did not exist already, then the developer may receive a `WARNING in unable to locate` message, and the file is not correctly copied. This can cause some weird behaviour, especially because if you run webpack again, it will work fine (because it's copied over the previously built file).

I would like to suggest adding in this documentation to the README to prevent people from using the plugin when it's not appropriate for their needs.

This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**


evilebottnawi:
Can you create minimum reproducible test repo, looks like a bug

soundasleep:
The minimum reproducible thing is:

- a webpack that compiles e.g. index.js -> dist/index.js
- configure copy-webpack-plugin to copy e.g. dist/index.js to dist/index2.js

The first time you run webpack, dist/index.js doesn't exist, so I get a warning that the file doesn't exist. The second time, as long as you haven't cleaned out dist/, the (old) file is copied fine.

evilebottnawi:
Please create minimum reproducible test repo, i can't reproduce you problem

evilebottnawi:
Sorry for delay, see what you mean, you let's merge this in docs

evilebottnawi:
Close in https://github.com/webpack-contrib/copy-webpack-plugin/pull/395, thanks for the PR

PR #373, opened 2019-04-22
fix to params that look like '[1]'

### LINKED ISSUE #369: toType detection has broken "to: '[1]'"
* Operating System: MacOS
* Node Version: 10.15.3
* NPM Version: 6.9.0
* webpack Version: 4.29.6
* copy-webpack-plugin Version: 5.0.2

### Expected Behavior

You should be able to specify a to: like '[1]' or 'foo/[1]'.

### Actual Behavior

Files are being put in a literal directory named '[1]'.

### Code

I believe this is caused somewhere in this change: https://github.com/webpack-contrib/copy-webpack-plugin/commit/52f8be6809ae0b876225973d34ff65a3b6009192

I apologize for not submitting a PR instead, as that is my preferred way to deal with bugs, but I do not have time to dig on this right now.

### How Do We Reproduce?

Create a to: setting of '[1]' that is a replacement from a RegExp test.

andyburke:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Fixes #369


andyburke:
I mean ... I pointed out the exact bug you guys caused in a refactor, you asked for a PR. I made a PR and now you ask for tests?

You make it pretty challenging to help you with your project.

evilebottnawi:
@andyburke without test we can break it again in future, so we need tests for this to avoid this sutiation

Issue #318, opened 2019-01-30
[Feature] Add order option


Scharnvirk:
Hello,

In a project I am developing, I have an issue with copying files with the same name from multiple sources.

Configuration of copy webpack plugin is as follows:
```
new CopyWebpackPlugin([
            { from: 'node_modules/engine/static', test: /\.(css|html|png)$/, force: true },
            { from: 'node_modules/subengine/static', test: /\.(css|html|png)$/, force: true },
            { from: './static', test: /\.(css|html|png)$/, force: true },
        ]),
```

The project has two dependencies, let's call them `engine` and `subengine`. They are both bundled within node_modules.

Each of these has its own graphical assets. In case the file with an exact same name and path is in more than one repo, then one from "project" is more important than one from `subengine`, and one from `subengine` is more important than file from `engine`. 

Let's say each of these contains an `abc.png` file in its own `static` folder. The intended result is to have `abc.png` come from projects's own `./static` folder, but it instead, the one from `subengine` is used.

HOWEVER.... which is where the fun begins...

If I do this:
```
new CopyWebpackPlugin([
            { from: 'node_modules/engine/static', test: /\.(css|html|png)$/, force: true },
            { from: 'node_modules/subengine/static', test: /\.(css|html|png)$/, force: true },
            { from: './static', test: /\.(css|html|png)$/, force: true },
        ], {debug: 'debug'}),
```

...then the order is correct, and `abc.png` is taken from project's `static` folder.

This might be random behavior and just a coincidence, but it really seems having debug on fixes the issue every time. 

I am using:
`"webpack": "~4.17.2",` (I would like not to update further for now because then the build breaks elsewhere)
`"copy-webpack-plugin": "^4.6.0",`

Build on Windows 10, Node 8.4.0

evilebottnawi:
> HOWEVER.... which is where the fun begins...

Just luck

Scharnvirk:
Thanks. 

It really seems consistent. We have automated tests and builds and it passes every time since I added the debug flag.

Is there any way to make this work "as intended" however? What other solution could I use if copy webpack plugin does not have ordering functionality?

evilebottnawi:
> Is there any way to make this work "as intended" however?

It is already work as intended.

> Is there any way to make this work "as intended" however? What other solution could I use if copy webpack plugin does not have ordering functionality?

Implement this. 
Package what can help with this https://github.com/sindresorhus/p-queue.
[...]

PR #374, opened 2019-04-24
refactor: make pattern processing order deterministic

MathijsMeijer:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
 
This change makes the pattern processing order deterministic, which allows users of this plugin to make assumptions about potentially conflicting assets paths. See #318 for an example use case. In my opinion, the potential performance regression this causes is fairly reasonable and secondary to the change. I think extending the API to manually control the pattern order, as suggested in #318, might make this unnecessarily complex. I'd love to hear your thoughts on this.

### Breaking Changes

n/a

### Additional Info

n/a


MathijsMeijer:
Thanks for reviewing and testing it @evilebottnawi. I’ve updated the pull request with a better approach which should make the impact on performance negligible, but I might be overseeing something. Are you willing to try it out?

MathijsMeijer:
All right. I’m not sure which ‘feature’ this code introduces though. I’ll close this PR; I’m probably better off writing a custom plugin since my requirements are increasingly deviating. Thanks.

PR #378, opened 2019-05-06
use posix path as assets key

snadn:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

should use posix path as assets key that consistent with webpack


evilebottnawi:
Why? What is problem?

snadn:
1. Consistent with webpack
image

2. Consistent with copy-webpack-plugin@4

3. Consistent with option and can be used by other plugin friendly
image

evilebottnawi:
Don't use `/` in to, it is not compatobility with windows, use `path.join`/`path.resolve`

snadn:
I think the key of assets should be treated as path of url instead of path of filesystem.

Regarding the third point, the value of 'to' may be 'js' (a dir), And my configuration in ‘tags’ is still expected to be ‘js/xxx.js’

And the first and second points are also important reasons, especially the first point.

evilebottnawi:
@snadn no, linux support `\` as part of name so in theory it can break cache

snadn:
@evilebottnawi But this is just that this pr modification is not good enough, but using ‘/’ is still more reasonable. Maybe getting posix path and record it when getting the file path will be better.

ps: just change the key of assets will break cache? Whether it is in window or linux, the keys obtained are the same.The potential problem should be that multiple files may be mapped to a key

evilebottnawi:
@snadn 
1. Use `normalize-path` instead `posix`
2. Need tests

snadn:
@evilebottnawi Can this pr be merged?

evilebottnawi:
Looks like related https://github.com/webpack-contrib/copy-webpack-plugin/issues/383

jeffposnick:
I was going to file a PR (see https://github.com/webpack-contrib/copy-webpack-plugin/compare/master...jeffposnick:windows-fix?expand=1) but I see that this PR should address it just as well.

PR #406, opened 2019-11-01
Performance improvement

sokra:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**
- [x] **performance upgrade**

### Motivation / Use-Case

Sets don't allow duplicates, so checking for `has` is not necessary

webpack 5 makes these Sets lazy, which makes reading more expensive.
So only the watcher should read from these Sets, which is called after the compilation has finished,
so that doesn't contribute to build performance.

LazySets also have a `addAll` method, which allows to add an Iterable lazily (only iterated when LazySet is read)
### Breaking Changes

No, webpack 4 compatibility is kept

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -98,28 +98,20 @@ class CopyPlugin {
     compiler.hooks.afterEmit.tapAsync(plugin, (compilation, callback) => {
       logger.debug('starting after-emit');
 
-      // Add file dependencies if they're not already tracked
-      for (const fileDependency of fileDependencies) {
-        if (compilation.fileDependencies.has(fileDependency)) {
-          logger.debug(
-            `not adding '${fileDependency}' to change tracking, because it's already tracked`
-          );
-        } else {
-          logger.debug(`adding '${fileDependency}' to change tracking`);
-
+      // Add file dependencies
+      if ("addAll" in compilation.fileDependencies) {
+        compilation.fileDependencies.addAll(fileDependencies);
+      } else {
+        for (const fileDependency of fileDependencies) {
           compilation.fileDependencies.add(fileDependency);
         }
       }
 
-      // Add context dependencies if they're not already tracked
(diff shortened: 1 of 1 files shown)


jquense:
This is breaking webpack4 for us. I think it's a bad interaction with some library using corejs, but interestingly if i log out `addAll` it's returning `function addAll() { [native code] }` which is a bit alarming. I cannot reproduce it in the repl tho!

FYI `addAll` is a proposed method for Set's so likely this test will fail if they are implemented https://github.com/tc39/proposal-collection-methods

evilebottnawi:
@jquense it is custom webpack class and not related to corejs

jquense:
@evilebottnawi I realize that, the problem is that it checks if `addAll` is in the class, in webpack4 it is a Set, and that webpack 4 Set also has `addAll` that works differently. because some other library (I think) uses corejs which is polyfilling `Set.addAll`

jquense:
In other words checking `'addAll' in compilation.fileDependencies` is not a safe check, because there is a proposal, and polyfill for a native `addAll` that works differently than `LazySet.addAll`. The code should probably duck type this differently

evilebottnawi:
@jquense what idea do you have to fix it? Maybe instance of solve this?

jquense:
either `instanceof` or maybe add a static symbol to LazySet so you can typecheck like `foo[Symbol.for('@@LazySet')] == true` or maybe just check the `constructor.name === 'LazySet'`?

evilebottnawi:
@jquense Can you send a PR?

jquense:
That would require i larger ecosystem change and i'm not sure what the desired approach would be. Generally happy to send PR's tho. @sokra should really be aware of this issue as well, b/c this test is likely in other places

PR #427, opened 2019-12-17
add concurrency control

porky-prince:
This PR contains a:

- [x] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Before modification, when too many files are copied, such as more than 20,000 files, "Error: EMFILE: too many open files, open.." will be generated. Even though I used 'graceful-fs', this error occurred when I had many other applications open on my computer, so I made a modification to limit the concurrency of fs, which was tested with no problem on Windows or Linux.

### Breaking Changes
No

### Additional Info
No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,16 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-### [5.1.2](https://github.com/webpack-contrib/copy-webpack-plugin/compare/v5.1.1...v5.1.2) (2019-12-15)
+### [5.1.2](https://github.com/webpack-contrib/copy-webpack-plugin/compare/v5.1.1...v5.1.2) (2019-12-17)
 
+
+### Features
+
+* add concurrency control
 
 ### Bug Fixes
 
-* fix error:too many open files
+* fix error: too many open files
 
 ### [5.1.1](https://github.com/webpack-contrib/copy-webpack-plugin/compare/v5.1.0...v5.1.1) (2019-12-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file. See [standa
 
 ### [5.1.2](https://github.com/webpack-contrib/copy-webpack-plugin/compare/v5.1.1...v5.1.2) (2019-12-15)
 
+
+### Bug Fixes
+
+* fix error:too many open files
+
 ### [5.1.1](https://github.com/webpack-contrib/copy-webpack-plugin/compare/v5.1.0...v5.1.1) (2019-12-12)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [5.1.2](https://github.com/webpack-contrib/copy-webpack-plugin/compare/v5.1.1...v5.1.2) (2019-12-15)
+
 ### [5.1.1](https://github.com/webpack-contrib/copy-webpack-plugin/compare/v5.1.0...v5.1.1) (2019-12-12)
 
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
...
+
+
+### Features
+
+* add concurrency control
+
+### Bug Fixes
+
+* fix error: too many open files
+
```
evilebottnawi: Avoid this, it is auto generated before we do release

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "copy-webpack-plugin",
-  "version": "5.1.1",
+  "version": "5.1.2",
```
evilebottnawi: Avoid it

src/index.js
```diff
@@ -77,10 +78,14 @@ class CopyPlugin {
                   return Promise.resolve();
                 }
 
+                const limit = pLimit(globalRef.concurrency || Infinity);
```
evilebottnawi: We already have this https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/processPattern.js#L25, maybe we should move it on top

CHANGELOG.md
porky-prince: ok,i got it.

package.json
porky-prince: ok.

src/index.js
porky-prince: I think we should keep it here:
```
const limit = pLimit(globalRef.concurrency || Infinity);
return Promise.all(
  files
    .filter(Boolean)
    .map((file) =>
      limit(() => postProcessPattern(globalRef, pattern, file))
    )
);
```
and remove  [ https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/processPattern.js#L25](https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/processPattern.js#L25)
Because the following code is a synchronous process and i don't understand why it's used here.
I'm confused.
```
        limit(() => {
          const file = {
            force: pattern.force,
            absoluteFrom: path.resolve(pattern.context, from),
          };

          file.relativeFrom = path.relative(pattern.context, file.absoluteFrom);

          if (pattern.flatten) {
            file.relativeFrom = path.basename(file.relativeFrom);
          }

          logger.debug(`found ${from}`);

          // Check the ignore list
          let il = pattern.ignore.length;

          // eslint-disable-next-line no-plusplus
          while (il--) {
            const ignoreGlob = pattern.ignore[il];

            let globParams = {
              dot: true,
              matchBase: true,
            };

            let glob;

            if (typeof ignoreGlob === 'string') {
              glob = ignoreGlob;
            } else if (isObject(ignoreGlob)) {
              glob = ignoreGlob.glob || '';

              const ignoreGlobParams = Object.assign({}, ignoreGlob);
              delete ignoreGlobParams.glob;

              // Overwrite minimatch defaults
              globParams = Object.assign(globParams, ignoreGlobParams);
            } else {
              glob = '';
            }

            logger.debug(`testing ${glob} against ${file.relativeFrom}`);

            if (minimatch(file.relativeFrom, glob, globParams)) {
              logger.info(
                `ignoring '${file.relativeFrom}', because it matches the ignore glob '${glob}'`
              );

              return Promise.resolve();
            }

            logger.debug(`${glob} doesn't match ${file.relativeFrom}`);
          }

          // Change the to path to be relative for webpack
          if (pattern.toType === 'dir') {
            file.webpackTo = path.join(pattern.to, file.relativeFrom);
          } else if (pattern.toType === 'file') {
            file.webpackTo = pattern.to || file.relativeFrom;
          } else if (pattern.toType === 'template') {
            file.webpackTo = pattern.to;
            file.webpackToRegExp = pattern.test;
          }

          if (path.isAbsolute(file.webpackTo)) {
            if (output === '/') {
              const message =
                'using older versions of webpack-dev-server, devServer.outputPath must be defined to write to absolute paths';

              logger.error(message);

              compilation.errors.push(new Error(message));
            }

            file.webpackTo = path.relative(output, file.webpackTo);
          }

          logger.info(
            `determined that '${from}' should write to '${file.webpackTo}'`
          );

          return file;
        })
```

evilebottnawi: Feel free to do it

porky-prince: > Feel free to do it

ok.

evilebottnawi:
Fixed in master, sorry for delay, anyway thanks for the PR

Issue #396, opened 2019-07-26
Retain file timestamps after copy


mcalmus:
- Operating System:
- Node Version:
- NPM Version:
- webpack Version:
- copy-webpack-plugin Version:

### Feature Proposal
See: https://github.com/angular/angular-cli/issues/14003
It would be nice if files copied using the plugin could retain their original timestamps rather than getting the current date and time. 

### Feature Use Case
When a build is done it becomes very difficult to determine what has changed and when transferring files to another system the timestamp can not be relied upon for quick checks.

evilebottnawi:
Please clarify what do you mean

mcalmus:
When files are copied using the plugin the standard OS file copy is used. The resulting file has a date and time of when the copy was performed. It would be nice if the original file's timestamp were retained.

For example:

Say I have a file:
>dir foo.png
02/21/2019  08:50         53,306 foo.png

When it's copied using the plugin it ends up as:
07/26/2019 10:00         53,306 foo.png

It would be nice if the copied file would retain the original February timestamp.

evilebottnawi:
keeping original date/time is invalid because copied files was created/modified at `07/26/2019`, i think it is unnecessary for this plugin, and very rare edge, you can create own plugin for this purpose

adrianhuna:
I got around this problem by using [rsync](https://linux.die.net/man/1/rsync)

```
rsync -a ~/path/to/project/src/assets/ ~/path/to/project/build/assets
```

akx:
We're using files copied by this plugin to upload a static site using `aws s3 sync`, which looks at file sizes and timestamps.
Currently all files get reuploaded on every sync, which is annoying.

I'll take a stab at implementing this, shouldn't be too hard.

setzamora:
Same case here with Azure Storage, it would be great if we can retain the timestamp of each file so it would prevent it from being re-uploaded. This would speed most CI/CD pipelines.

evilebottnawi:
Feel free to send a PR

akx:
> Feel free to send a PR

I have an open branch about this, but I got sidetracked. Perhaps I'll get something complete done this week or the next!

DYefimov:
Found another problem with this same root cause:
Chrome devtools workspaces 2.0 is not mapping to local resources correctly when files differ by timestamp.
As a result: webpack bundled code is editable from devtools, where static files are not.

Would be nice to have this option for development env.
IMHO this option should be true by default.

[...]

PR #435, opened 2020-03-09
"Keep times" option

akx:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

This adds the "keep file times" functionality discussed in #396.

### Breaking Changes

This functionality is opt-in, so nothing should break (except maybe if this plugin is used in a web context without a shim for `fs`, see below).

### Additional Info

The code in `updateTimes` is a little finicky since we need to find an implementation of `utimes` to use. When run in a web context (or with an output file system we don't support), turning `keepTimes` on will just result in a bunch of warnings.

According to https://github.com/webpack/webpack/commit/e9c0d068ddbb3213e7601138c5403598da6ac779 the default output filesystem in the forthcoming Webpack 5 will just be `graceful-fs`, which has `utimes` natively. If the output filesystem object has `utimes`, we expect it to be a function that smells like `fs.utimes()`. (This will work for any custom filesystem object too.)

For Webpack 4, however, the default output file system in Webpack 4 is a custom `NodeOutputFileSystem` which lacks `utimes` altogether. However, since it is [a simple wrapper around `fs`](https://github.com/webpack/webpack/blob/5f65ecbf1999ebe82259927289073a285b0e8a6c/lib/node/NodeOutputFileSystem.js), we can assume that `fs.utimes()` works fine for assets output with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -505,6 +505,7 @@ module.exports = {
 |         [`ignore`](#ignore)         |  `{Array}`  |            `[]`            | Array of globs to ignore (applied to `from`)                                                                                                      |
 |        [`context`](#context)        | `{String}`  | `compiler.options.context` | A path that determines how to interpret the `from` path, shared for all patterns                                                                  |
 | [`copyUnmodified`](#copyunmodified) | `{Boolean}` |          `false`           | Copies files, regardless of modification when using watch or `webpack-dev-server`. All files are copied on first build, regardless of this option |
+|      [`keepTimes`](#keeptimes)      | `{Boolean}` |          `false`           | Copy the original access and modification over to the destination files, when possible                                                            |
 
 #### `logLevel`
 
@@ -565,6 +566,18 @@ Copies files, regardless of modification when using watch or `webpack-dev-server
 ```js
 module.exports = {
   plugins: [new CopyPlugin([...patterns], { copyUnmodified: true })],
+};
+```
+
+#### `keepTimes`
+
+Attempt to copy source files' access and modification times over to the destination files.
+
+**webpack.config.js**
+
+```js
+module.exports = {
+  plugins: [new CopyPlugin([...patterns], { keepTimes: true })],
 };
 ```
 
--- a/src/index.js
+++ b/src/index.js
@@ -7,6 +7,7 @@ import schema from './options.json';
 import preProcessPattern from './preProcessPattern';
 import processPattern from './processPattern';
 import postProcessPattern from './postProcessPattern';
+import updateTimes from './updateTimes';
 
 class CopyPlugin {
   constructor(patterns = [], options = {}) {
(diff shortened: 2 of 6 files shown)


evilebottnawi:
@akx Do you still need this feature? I can write full roadmap for correct implementation

akx:
> @akx Do you still need this feature? I can write full roadmap for correct implementation 

Sure. This implementation seems to work though, it's just the npm audit snafus keeping it back. What's wrong with the implementation in your opinion?

evilebottnawi:
@akx sorry, for delay, it should be solved on webpack side:

- we use `webpack-sources` for assets, so we need implement property `stats()` for https://github.com/webpack/webpack-sources/blob/master/lib/Source.js#L7
- emit `stats` from `source` here https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L546 and write stats to saved file
- implement `stats` keeping here

In the future, it will also allow other plugins modify stats

akx:
@evilebottnawi 

So... that means changes to both `webpack-sources` and `webpack` core itself, with the uncertainty of whether or not they'll get approved at all to begin with? Also, considering these are API changes, they'd probably be version-major?

If I rebase this PR, would it be approved until `stats` collection becomes (if it ever will be) implemented in Webpack itself?

evilebottnawi:
We can't implement it on plugin side, we use `compilation.hooks.additionalAssets` hook, and on this hook assets do not exist yet on disk, they exit only on `compilation.hooks.assetEmitted` hook and on it we cannot modify assets, It’s also worth saying that we don’t know in what environment the webpack works, so `fs` module can't exists (I prepare a PR to use webpack input file-system for `fast-glob` so CopyWebpackPlugin can work in browsers without any problems). Therefore, this needs to be improved on the side of the webpack itself, and not the plugin.

We already had a discussion about this, and it was decided that the implementation on webpack-sources side it the best solution. Why we still haven’t done it? Because this is a rare request.

No need to be afraid to improve something.

akx:
> they exit only on `compilation.hooks.assetEmitted` hook and on it we cannot modify assets

We're only modifying filesystem metadata that Webpack doesn't care about. That sounds like it should be safe.

>, It’s also worth saying that we don’t know in what environment the webpack works, so `fs` module can't exists

The PR is aware of this and does a best effort to keep times: https://github.com/webpack-contrib/copy-webpack-plugin/pull/435/files#diff-2e0f2c5a9d0afaccd39535254a1c9076R1-R20

evilebottnawi:
This implementation unsafe and potential break webpack-dev-server, solution is above

akx:
I guess I'll just close this...

PR #437, opened 2020-03-09
Validate global options too

akx:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I noticed #419 added validation for the `patterns` array, but the global options (which are touched by e.g. #435) aren't validated. (I may or may not have banged my head against the wall there.)

The documentation was also updated to reflect the fact that one can actually pass in Minimatch settings objects for ignore too.

### Breaking Changes

This may break some downstream software misusing the `options` object.

### Additional Info

"Test update" is not checked, because tests don't pass in naughty options.

### DIFF
--- a/src/options.json
+++ b/src/options.json
@@ -1,90 +1,33 @@
 {
-  "definitions": {
-    "ObjectPattern": {
-      "type": "object",
-      "properties": {
-        "from": {
-          "anyOf": [
-            {
-              "type": "string",
-              "minLength": 1
-            },
-            {
-              "type": "object"
-            }
-          ]
-        },
-        "to": {
-          "type": "string"
-        },
-        "context": {
-          "type": "string"
-        },
-        "toType": {
-          "enum": ["dir", "file", "template"]
-        },
-        "test": {
-          "anyOf": [
-            {
-              "type": "string"
-            },
-            {
-              "instanceof": "RegExp"
-            }
-          ]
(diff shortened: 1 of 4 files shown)


src/options.json
```diff
@@ -1,90 +1,34 @@
...
-    "anyOf": [
-      {
-        "$ref": "#/definitions/StringPattern"
-      },
-      {
-        "$ref": "#/definitions/ObjectPattern"
-      }
-    ]
-  }
+  "required": [],
```
evilebottnawi: Just remove it

evilebottnawi:
Can you update jest to avoid npm problem too?

akx:
> Can you update jest to avoid npm problem too?

@evilebottnawi No, that was discussed in the comment in #438 – updating Jest (via `npm audit --fix`) breaks compatibility with Node 6 and Node 8 since it begins using native `async`/`await`.

evilebottnawi:
Thanks for the PR, we will remove global options in the next major release, because we already remove `logLevel` (we use built-in webpack logger you can setup level using webpack config) and `copyUnmodifed` options (it was bad design we need persist assets between builds), so I am closing this PR, I start to removing global options support

PR #438, opened 2020-03-10
chore: fix npm audit woes

akx:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Azure Pipelines' audit step has been broken for a while. 

This PR relaxes the `security` step:

* only check production dependencies
 * don't have low-level vulnerabilities break the build
* output parseable format to avoid huge ANSI art log

### Breaking Changes

None.

PR #441, opened 2020-03-30
Update Jest dependency + drop testing on Node 6

akx:
This PR contains a:

- [x] **test update**

### Motivation / Use-Case

CI had been broken due to npm audit problems. This should fix it.
However, the new version of Jest here isn't compatible with Node 6, so testing on it must be dropped.

In addition, it took some package-lock hacking to fix the one remaining low-priority problem in optimist (which might be fixed any day now, see https://github.com/wycats/handlebars.js/issues/1661)...

Closes #438
Closes #439

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -55,9 +55,6 @@ jobs:
         node-8:
           node_version: ^8.9.0
           webpack_version: latest
-        node-6:
-          node_version: ^6.9.0
-          webpack_version: latest
         node-10-canary:
           node_version: ^10.13.0
           webpack_version: next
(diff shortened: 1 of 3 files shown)


akx:
@evilebottnawi Is this acceptable (as opposed to https://github.com/webpack-contrib/copy-webpack-plugin/pull/438#discussion_r395555739 and so on)?

If it is, someone will need to turn off the requirement for node-6 checks in the repo's settings.

akx:
> It is breaking change, we can't merge it, just ignore npm audit

Fair – that's basically what I did in #438. Audit is part of the lint step, which is _also_ a required check. Would you like me to fully remove the audit from the lint step?

evilebottnawi:
Just ignore it, problems with audit will be resolved in near future

evilebottnawi:
Done in master, thanks for the PR, ETA - next week

PR #472, opened 2020-05-14
chore: drop test option in favor transformPath

cap-Bernardito:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

Drop `test` option in favor `transformPath` option

### Breaking Changes

Yes

### Additional Info

No


aaarichter:
Hello @evilebottnawi & @cap-Bernardito,
 
I do feel that this change is a step back. The `test` option was simple and self explanatory vs 
```js
 transformPath(targetPath, absolutePath) {
  const matches = absolutePath.match(/\.([^.]*)$/);
  const [, res] = matches;
  const target = targetPath.replace(/\[[1]\]/, res);

  return target;
},
```
What was the reason to drop the `test` option?

The [Readme](https://github.com/webpack-contrib/copy-webpack-plugin#transformpath) does not have an example on how to use `transformPath` to replace `test`.

evilebottnawi:
@aaarichter `test` was buggy and never works as expected + problems on windows, I don't know your logic, you have a function so you can create any logic

PR #507, opened 2020-06-18
fix: do not execute on a child compilation

### LINKED ISSUE #505: In 6.x files are transformed/copied multiple times per build
- Operating System: MacOS 10.15.5 (19F101)
- Node Version: v14.4.0
- NPM Version: N/A (Yarn 1.22.4)
- webpack Version: webpack@4.43.0
- copy-webpack-plugin Version: 6.x

### Expected Behavior

Files should be transformed/copied only once per build

### Actual Behavior

In 5.x the files are transformed once per build (verified by logging when `transform` is called). In 6.x the `transform` function is called a large number of times (perhaps 6-10 times per file.

I noticed that removing `new VueLoaderPlugin()` from the plugins list reduced the multiple calls down to two per file. My suspicion is that in 6.x you're hooking into an event that runs even on child builds rather than on the entry point build. `new VueLoaderPlugin()` seems to kick off a lot of child builds depending on the number of Vue components in the application.

Reverting to 5.x fixed the issue.

### Code

```js
// webpack.config.js

...
transform(content, filePath) {
  console.log(filePath)
}
...
```

```js
new VueLoaderPlugin()
```

### How Do We Reproduce?
The only hint I have is to do with `new VueLoaderPlugin()`. Hopefully this is enough of a pointer, but if not, let me know and I'll look into creating a minimal reproducible.

evilebottnawi:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #505

### Breaking Changes

No

### Additional Info

No


alan-agius4:
@evilebottnawi, is there any chance of getting a release containing this fix cut soon? This is effecting us (the Angular team) in a bad way see https://github.com/angular/angular-cli/issues/18087

evilebottnawi:
@alan-agius4 Yep, let's do it, want to solve https://github.com/webpack-contrib/copy-webpack-plugin/issues/504 also in this patch, but it is not critical

evilebottnawi:
@alan-agius4 Done https://github.com/webpack-contrib/copy-webpack-plugin/releases/tag/v6.0.3

alan-agius4:
Thanks @evilebottnawi

PR #541, opened 2020-10-21
feat: persistent cache between runs (webpack@5 only)

### LINKED ISSUE #504: After each code change, the copy operation of the "copy-webpack-plugin" plug-in is executed at the same time as the recompilation operation.
webpack: v4.43.0
webpack-dev-server: v3.11.0
copy-webpack-plugin: v6.0.2
html-webpack-plugin v4.3.0
Windows 10 professional edition 1902 18363.900
node.js: v14.4.0
npm: v6.14.5

Describe the bug:
1、
After each code change, the copy operation of the "copy-webpack-plugin" plug-in is executed at the same time as the recompilation operation.
image

To Reproduce:
1、The configuration of "copy-webpack-plugin" is:
image
2、The configuration of "html-webpack-plugin" is:
image
3、The configuration of "webpack" is:
image

Expected behavior：
1、
After each code change, the recompile operation will not be executed at the same time as the "copy-webpack-plugin" plugin's copy operation.
2、
The "copy-webpack-plugin" plug-in will be re copied only when the copied file changes.

---

This issue was moved from webpack/webpack#11032 by @evilebottnawi. Original issue was by @LP1994.

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

fixes #504 

### Breaking Changes

No

### Additional Info

only webpack@5


src/index.js
```diff
@@ -37,8 +37,57 @@ class CopyPlugin {
...
+    }
+
+    // eslint-disable-next-line consistent-return
+    return new Promise((resolve, reject) => {
+      compilation.fileSystemInfo.createSnapshot(
+        // eslint-disable-next-line no-undefined
+        undefined,
+        [dependency],
+        [],
+        [],
```
sokra, line 52: You can pass `undefined` here, that saves a bit memory.

src/index.js
```diff
@@ -266,64 +314,129 @@ class CopyPlugin {
...
+                compilation.errors.push(error);
 
-              logger.debug(
-                `getting cached transformation for "${file.absoluteFrom}"`
-              );
+                return;
+              }
+            }
+          }
+        }
```
sokra, line 361: `checkSnapshotValid` on a fresh created snapshot will always be true. You need to call `checkSnapshotValid` on a snapshot that was created in the last run of webpack. The snapshot need to be stored and restored in and from the cache.

Here is some pseudo code:

``` js
// 1. check cache first, validate snapshot in cache entry
cacheEntry = await cache.getPromise(file.relativeFrom, null);
if (cacheEntry !== undefined) {
  const { snapshot, source } = cacheEntry;
  if (await checkSnapshotValid(snapshot)) {
    // return early when cache entry can be used
    return source;
  }
}
// 2. create a new snapshot
const snapshot = await createSnapshot(file.relativeFrom);
// 3. create new data
const source = await createNewSource(file.relativeFrom);
// 4. store both in cache
await cache.storePromise(file.relativeFrom, null, { source, snapshot });
// 5. return new data
return source;
```

You are in the good position that you can call 2. before 3. Usually that's not possible, because dependencies are not known before execution. In that case one would do this:

``` js
// 2a.
const startTime = Date.now();
// 2b. create new data
const source = await createNewSource(file.relativeFrom);
// 3. create a new snapshot
const snapshot = await createSnapshot(file.relativeFrom, startTime);
```

src/index.js
```diff
@@ -37,8 +37,57 @@ class CopyPlugin {
...
+    });
+  }
+
+  static async checkSnapshotValid(compilation, snapshot) {
+    if (!compilation.fileSystemInfo) {
+      return;
+    }
+
+    // eslint-disable-next-line consistent-return
+    return new Promise((resolve, reject) => {
```
sokra, line 73: https://nodejs.org/api/util.html#util_util_promisify_original

```  js
const checkSnapshotValid = util.promisify((compilation, snapshot, callback) =>
  compilation.fileSystemInfo.checkSnapshotValid(snapshot, callback)
);
const createSnapshot = util.promisify((compilation, dependency, callback) =>
  compilation.fileSystemInfo.createSnapshot(Date.now(), [dependency], undefined, undefined, null, callback)
);
```

sokra: Passing an `undefined` startTime, would assume worse case.

sokra (changes requested):

PR #590, opened 2021-03-03
Feat: added `priority` option

### LINKED ISSUE #318: [Feature] Add order option

cap-Bernardito:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

Fixed #318 

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ module.exports = {
 |           [`filter`](#filter)           |     `{Function}`     |                   `undefined`                   | Allows to filter copied assets.                                                                                                                        |
 |           [`toType`](#totype)           |      `{String}`      |                   `undefined`                   | Determinate what is `to` option - directory, file or template.                                                                                         |
 |            [`force`](#force)            |     `{Boolean}`      |                     `false`                     | Overwrites files already in `compilation.assets` (usually added by other plugins/loaders).                                                             |
+|         [`priority`](#priority)         |      `{Number}`      |                       `0`                       | Allows you to specify the copy priority.                                                                                                               |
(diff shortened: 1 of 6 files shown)


alexander-akait (approved):

Eutrepe:
Thank you for this option <3

PR #645, opened 2021-11-23
feat: implement support for array values in `from` property. Closes #631

### LINKED ISSUE #631: Allow `from` to be array of paths / glob, respect order
### Feature Proposal

Please allow the **from** property to contain an array of paths / glob patterns
because I need to include files from several different locations and copy them all into one folder.
This is quite ugly and verbose doing it with glob.

The assets array in transformAll should then be returned **in the order of the from array** so you can rely on javascript execution order.

### Feature Use Case
Copying and or concatenating and transforming (javascript) files from multiple locations into one file / folder.

### Please paste the results of `npx webpack-cli info` here, and mention other relevant information
not relevant

orangy:
This PR contains an implementation for processing arrays in `from` property, and is:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

In my project I'm missing `transformAll` on a bunch of files, so I've found a recent issue #631 and decided to implement it. 

### Breaking Changes

There shouldn't be any breaking changes since it doesn't disable or break any scenario that was possible before, and only creates new (wonderful!) possibility to make copy scripts shorter and more powerful.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -633,15 +633,41 @@ class CopyPlugin {
               let assets;
 
               try {
-                assets = await CopyPlugin.runPattern(
-                  globby,
-                  compiler,
-                  compilation,
-                  logger,
-                  cache,
-                  item,
-                  index
-                );
+                if (item.from instanceof Array) {
+                  if (!item.from.every((from) => typeof from === "string")) {
+                    compilation.errors.push(
+                      new Error(
+                        `Invalid "pattern.from": ${item.from}, every element should be a string"`
+                      )
+                    );
+                  }
+
+                  assets = [].concat(
+                    ...(await Promise.all(
+                      item.from.map(async (from) =>
+                        CopyPlugin.runPattern(
+                          globby,
+                          compiler,
+                          compilation,
+                          logger,
+                          cache,
+                          { ...item, from },
+                          index
+                        )
+                      )
+                    ))
+                  );
+                } else {
+                  assets = await CopyPlugin.runPattern(
+                    globby,
+                    compiler,
+                    compilation,
+                    logger,
+                    cache,
+                    item,
+                    index
+                  );
+                }
               } catch (error) {
                 compilation.errors.push(error);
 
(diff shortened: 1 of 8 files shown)


src/index.js
```diff
@@ -633,15 +633,41 @@ class CopyPlugin {
...
-                  index
-                );
+                if (item.from instanceof Array) {
+                  if (!item.from.every((from) => typeof from === "string")) {
+                    compilation.errors.push(
+                      new Error(
+                        `Invalid "pattern.from": ${item.from}, every element should be a string"`
+                      )
+                    );
+                  }
```
alexander-akait, line 643: Avoid it, we validate it in schema

src/index.js
```diff
@@ -633,15 +633,41 @@ class CopyPlugin {
...
+                          compiler,
+                          compilation,
+                          logger,
+                          cache,
+                          { ...item, from },
+                          index
+                        )
+                      )
+                    ))
+                  );
```
alexander-akait, line 659: Why not pass array to globby? They support it https://github.com/sindresorhus/globby#usage

src/options.json
```diff
@@ -5,13 +5,25 @@
...
+              },
+              "minItems": 1
+            }
+          ],
+          "description": "Glob or path from where we copy files, or array of paths or globs.",
+          "link": "https://github.com/webpack-contrib/copy-webpack-plugin#from"
         },
         "to": {
-          "anyOf": [
+          "oneOf": [
```
alexander-akait, line 26: Why it was changed?

alexander-akait (changes requested):
Some notes

jasper2virtual:
Hi, I need this feature too, I notice this PR is stopped for a year, what can I do or contribute to make it go?

alexander-akait:
Yes, feel free to send a new PR

Issue #788, opened 2024-10-26
Switch to tinyglobby


benmccann:
### Modification Proposal

Switch from `globby` to `tinyglobby`.

`tinyglobby` has quickly been [adopted by `vite`](https://github.com/vitejs/vite/blob/89f8303e727791aa7be6f35833a708b6a50e9120/packages/vite/package.json#L151) and [a number of other major projects](https://www.npmjs.com/package/tinyglobby) as a much smaller near drop-in replacement for `globby`.

The main challenge with adopting it in this library is that the `globby` constructor options are directly exposed as an option here and the `tinyglobby` constructor options type is different, so it would be a breaking change that would have to be done in a major release.

### Expected Behavior / Situation

https://npmgraph.js.org/?q=tinyglobby - 2 dependencies

### Actual Behavior / Situation

https://npmgraph.js.org/?q=globby - 23 dependencies

### Please paste the results of `npx webpack-cli info` here, and mention other relevant information

n/a

evenstensberg:
Sounds good. I also think we can remove external libraries for globbing once we reach to the point that the lowest nodejs version supported allows us to use the native node glob.

Send a PR!

benmccann:
I've made a lot of progress on this. It seems that `absolute: true` needs to be set. It appears that `globby` has some bug that `tinyglobby` doesn't because the default is `false`, but `globby` seems to be returning absolute paths anyway. We need to set `absolute: true` to have `tinyglobby` also return absolute paths. Nearly all the tests pass once that's done. There's still one last issue I have to investigate

alexander-akait:
@benmccann Feel free to feedback if you will have questions

benmccann:
This might be blocked by `tinyglobby` adding support for the `fs` option: https://github.com/SuperchupuDev/tinyglobby/issues/78

alexander-akait:
We can add it without this option and check perd

benmccann:
Okay, great! I will fix the last test and send a PR then

Juraj-Masiar:
Hello,
I'm trying to upgrade to new version 13.0.0 but it's failing due to this breaking change and the text in the [release notes ](https://github.com/webpack-contrib/copy-webpack-plugin/releases/tag/v13.0.0) is not clear enough about what to do actually.
This is my super simple usecase:
```ts
new CopyPlugin({
  patterns: [
    { from: browser,  to: '', globOptions: { ignore: ['**.ts', '**.git'] } },
[...]

PR #789, opened 2024-10-27
chore: simplify glob usage

benmccann:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I started looking at implementing https://github.com/webpack-contrib/copy-webpack-plugin/issues/788 and noticed that some of the glob usage here was more complicated than it needed to be.

The first thing that was being done with the glob results was to check if it's a file. However, globby has an `onlyFiles` option that we can use to avoid the need for that check. After that's done, there's no longer a need to set `objectMode`. Thus, instead of getting back an object, we can simply receive back a string, which is both simpler and better for performance.

### Breaking Changes

n/a

### Additional Info

I may follow up with another PR to switch to `tinyglobby`. However, I thought it was better to separate those two refactorings to keep the PRs small and make it easier to debug if there are any unexpected issues

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -340,9 +340,11 @@ class CopyPlugin {
 
     /** @type {GlobbyOptions & { objectMode: true }} */
     const globOptions = {
-      ...{ followSymbolicLinks: true },
+      followSymbolicLinks: true,
       ...(pattern.globOptions || {}),
-      ...{ cwd: pattern.context, objectMode: true },
+      cwd: pattern.context,
+      objectMode: true,
+      onlyFiles: true,
     };
 
     // @ts-ignore
@@ -412,7 +414,9 @@ class CopyPlugin {
     let globEntries;
 
     try {
-      globEntries = await globby(glob, globOptions);
+      globEntries = globOptions.onlyDirectories
+        ? []
+        : await globby(glob, globOptions);
     } catch (error) {
       compilation.errors.push(/** @type {WebpackError} */ (error));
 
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -43,7 +43,6 @@ const getGlobby = memoize(async () => {
 /** @typedef {import("webpack").WebpackError} WebpackError */
 /** @typedef {import("webpack").Asset} Asset */
 /** @typedef {import("globby").Options} GlobbyOptions */
-/** @typedef {import("globby").GlobEntry} GlobEntry */
 /** @typedef {ReturnType<Compilation["getLogger"]>} WebpackLogger */
```
evenstensberg, line 46: remove space

evenstensberg (commented):
I've enabled the CI's. If some tests fail we need to address those. I also want to add some more tests that verifies the changes.

src/index.js
```diff
@@ -407,12 +408,14 @@ class CopyPlugin {
     logger.log(`begin globbing '${glob}'...`);
 
     /**
-     * @type {GlobEntry[]}
+     * @type {string[]}
```
evenstensberg, line 411: whats the typedef change here?

benmccann: `objectMode: true` changes the return type from the default `string[]` to `GlobEntry[]`. Since we've removed `objectMode: true` we can go back to `string[]`, which is simpler and has better performance

src/index.js
```diff
@@ -407,12 +408,14 @@ class CopyPlugin {
...
      */
     let globEntries;
 
     try {
-      globEntries = await globby(glob, globOptions);
+      globEntries = globOptions.onlyDirectories
```
alexander-akait, line 416: Let's avoid it and just add the `onlyDirectories` doesn't supported in documentation. It doesn't make sense for our plugin anyway.

And we will still make a major release after this pr

PR #790, opened 2024-11-02
chore: upgrade cspell

benmccann:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

adds `tinyglobby` to the included dictionary, which is necessary for https://github.com/webpack-contrib/copy-webpack-plugin/issues/788

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,7 @@
         "@webpack-contrib/eslint-config-webpack": "^3.0.0",
         "babel-jest": "^29.7.0",
         "cross-env": "^7.0.3",
-        "cspell": "^8.13.1",
+        "cspell": "^8.15.6",
         "del": "^6.1.1",
         "del-cli": "^5.1.0",
         "eslint": "^8.57.0",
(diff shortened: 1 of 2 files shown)


snitin315:
> adds tinyglobby to the included dictionary,

How is this related to #788? Can't we just add `tinyglobby` [here](https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/.cspell.json) during migration?

benmccann:
> How is this related to https://github.com/webpack-contrib/copy-webpack-plugin/issues/788?

The word `tinyglobby` must be in the dictionary or the lint job for spell checking will fail. The latest version of cspell has it included 

> Can't we just add tinyglobby [here](https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/.cspell.json) during migration?

That would probably work too. Is it simpler than this solution though? Presumably we'll update to the latest cspell eventually, so why not do that alone instead of doing it and also maintaining it in a separate list?

alexander-akait (approved):

PR #791, opened 2024-12-02
refactor: switch from fast-glob to tinyglobby

benmccann:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Ref https://github.com/webpack-contrib/copy-webpack-plugin/issues/788

### Breaking Changes

None

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,12 +9,12 @@
       "version": "12.0.2",
       "license": "MIT",
       "dependencies": {
-        "fast-glob": "^3.3.2",
         "glob-parent": "^6.0.1",
         "globby": "^14.0.1",
         "normalize-path": "^3.0.0",
         "schema-utils": "^4.2.0",
-        "serialize-javascript": "^6.0.2"
+        "serialize-javascript": "^6.0.2",
+        "tinyglobby": "^0.2.10"
       },
       "devDependencies": {
         "@babel/cli": "^7.24.6",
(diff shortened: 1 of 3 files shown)


alexander-akait (approved):
You look good, have you tried to measure their performance?

benmccann:
@alexander-akait can you share the benchmark you ran?

The only thing I'm switching here at the moment is which `escapePath` method is used. Is that the method you benchmarked? `tinyglobby`'s implementation is extremely simple as all it does is a regex replacement:
https://github.com/SuperchupuDev/tinyglobby/blob/d51e2b98d18486a904193e100e22a9c9ed8cbb08/src/utils.ts#L31

I would think that this method should be able to be run many orders of magnitude faster than that and any performance difference would just be noise in the overall performance of `copy-web-plugin` as this is such an infinitesimally small portion of the overall work that the plugin is doing

alexander-akait:
@benmccann Oh, I see, I just compared the speed of work between `tinyglobby` and `fast-glob`, as I understand it, this pr is not finished yet?

benmccann:
I thought I would break up the work to make things more reviewable and lower risk. This PR switches out all usage of fast-glob for tinyglobby, which I think we could merge here and then I could follow up with another PR to switch out globby for tinyglobby as well. If you're really worried about performance I could wait until the next tinyglobby release to do the second half as they're currently working to make sure tinyglobby is always faster than alternatives in https://github.com/SuperchupuDev/tinyglobby/pull/76

alexander-akait:
@benmccann let's put all in one pr place, I don't think there will be many of them

benmccann:
Okay. I'll close this for the time being then

alexander-akait:
@benmccann We can keep open, no problems

PR #825, opened 2026-04-03
Fix/concurrency

nandini888-hash:
. Concurrency Optimization
Improved handling of concurrency handling using cpu-based dynamic scaling
Better performance when processing multiple files

In copy-webpack-plugin/src/index.js


alexander-akait (changes requested):
This option is not related to CPU at all, it is about fs calls, your code makes the plugin is slower

PR #826, opened 2026-04-04
feat: allow from option to be an array of paths

### LINKED ISSUE #631: Allow `from` to be array of paths / glob, respect order

CoderSerio:
**Summary**

Fixes #631

**What kind of change does this PR introduce?**

feat:

- Support copying multiple files from array, and as the same time, maintain the order of files from array
- Add test file for the feature

**Did you add tests for your changes?**

Yes :)

**Does this PR introduce a breaking change?**

No, it's fully backward compatibility

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

- Extend the documented type of from from string to string | string[]
- From now supports copying multiple files from an array
- 
```js
new CopyPlugin({
  patterns: [
    {
      from: ["LICENSE", "README.md", "package.json"],
      to: "dist/",
    },
  ],
});
```

**Use of AI**

I've read the [AI policy](https://github.com/webpack/governance/blob/main/AI_POLICY.md), and been clear totally about my code, and there are steps I used AI to collaborate:

1. Confirm the original behavior of the from config field to make sure there was no misunderstanding
2. Speed up test-driven development by generating initial test cases（I've reviewd them all)
3. Review whether the cache index design was reasonable and validate the logic，and it proved that in math (that's cool, and the thing I could not do the same

```md
The previous additive encoding strategy (`p + i`) can cause collisions because different `(p, i)` pairs may produce the same result, such as `(0, 5)` and `(5, 0)`.

The new encoding strategy (`K * p + i`), with the constraint `0 <= i < K`, is collision-free within the valid range. If `K * p1 + i1 = K * p2 + i2`, then `K * (p1 - p2) = i2 - i1`; since the right-hand side must be smaller than `K` in absolute value while also being divisible by `K`, it can only be `0`, which means `p1 = p2` and `i1 = i2`.

Intuitively, this works like combining the pattern index and the file index as digits in a mixed-radix number system. As long as `i` never reaches `K`, the lower part can never overflow into the higher part. Human beings reinventing positional notation to avoid cache collisions. Civilization limps onward.

```

4. Perform a basic review of the code to help identify any obvious bugs

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -874,22 +874,67 @@ class CopyPlugin {
                */
               let copiedResult;
 
-              try {
-                copiedResult = await CopyPlugin.glob(
-                  globby,
-                  compiler,
-                  compilation,
-                  logger,
-                  cache,
-                  concurrency,
-                  /** @type {ObjectPattern & { context: string }} */
-                  (pattern),
-                  index,
-                );
-              } catch (error) {
-                compilation.errors.push(/** @type {Error} */ (error));
+              const fromList = Array.isArray(pattern.from)
+                ? pattern.from
+                : [pattern.from];
 
-                return;
+              if (fromList.length === 0) {
+                copiedResult = [];
+              } else if (fromList.length > 1) {
+                const results = [];
+
+                for (let i = 0; i < fromList.length; i++) {
+                  const from = fromList[i];
+                  const arrayPattern = { ...pattern, from };
+
+                  try {
+                    const result = await CopyPlugin.glob(
+                      globby,
+                      compiler,
+                      compilation,
+                      logger,
+                      cache,
+                      concurrency,
+                      /** @type {ObjectPattern & { context: string }} */
+                      (arrayPattern),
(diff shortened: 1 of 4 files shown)


CoderSerio:
👀 Hello, is there any possibility to merge this feature？

alexander-akait (changes requested):
Sorry we don't need it, you can pass many `from` already

CoderSerio:
> Sorry we don't need it, you can pass many `from` already

I see, thanks for your reply. I'll close this pr.

